PR #1, opened 2021-07-26
Learning APIs by getting fire data

kyleredilla:
This PR enables the API to fetch the fire data for a given lat and lon.

### DIFF
--- a/application.py
+++ b/application.py
@@ -165,17 +165,14 @@ def package_landcover(landcover_response):
 
 
 def validate_latlon(lat, lon):
-    """Validate the lat and lon values, 
+    """Validate the lat and lon values,
     return bool for validity"""
     try:
         lat_numeric = isinstance(int(float(lat)), int) or isinstance(float(lat), float)
         lon_numeric = isinstance(int(float(lon)), int) or isinstance(float(lon), float)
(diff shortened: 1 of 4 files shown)


application.py
```diff
@@ -1,15 +1,203 @@
...
+        landcover_package["type"] = landcover_names[code]["type"]
+        landcover_package["color"] = landcover_names[code]["color"]
+    return landcover_package
+
+
+def validate_latlon(lat, lon):
+    """Validate the lat and lon values,
+    return bool for validity"""
+    try:
+        lat_numeric = isinstance(int(float(lat)), int) or isinstance(float(lat), float)
```
kyleredilla, line 171: @charparr I'm a little confused about these lines. Shouldn't these always return true if there is no error in the initial conversion from float() or int()?

charparr: These two lines are left over from testing the code and should be deleted. The inequalities below call `float()` within the `try`:. That should ensure it is either float, int, or string that is convert-able, and so should be able to be placed in the URL string.

charparr (changes requested):

application.py
kyleredilla: commit added to remove these.

charparr (approved):

PR #2, opened 2021-07-29
Merge "Split api 2" down to main

charparr:
This PR adds the rest of the permafrost endpoints for the permafrost data in Geoserver (and the map playground) and implements a Blueprint pattern where all specific routes live in their own directory.


routes/permafrost.py
```diff
@@ -0,0 +1,152 @@
...
+    if obu_vector_resp["features"] == []:
+        obu_vector_pkg["Obu vector data"] = "No data at this location."
+    else:
+        obu_vector_pkg["Permafrost Extent"] = obu_vector_resp["features"][0][
+            "properties"
+        ]["PFEXTENT"]
+    return obu_vector_pkg
+
+
+@routes.route("/🧊/<lat>/<lon>")
```
brucecrevensten: As much as it pains me to be serious, we probably should make this one use `permafrost` as an endpoint.

routes/fire.py
```diff
@@ -0,0 +1,178 @@
+import asyncio
+from aiohttp import ClientSession
+from flask import abort, Blueprint, render_template
+from validate_latlon import validate
+from . import routes
```
brucecrevensten, line 5: What exactly is this line doing?  Pulling in `routes` from `__init__`, ...?  Should I just read up on this?

routes/fire.py
```diff
@@ -0,0 +1,178 @@
...
+    1: "#2b83ba",
+    2: "#abdda4",
+    3: "#ffffbf",
+    4: "#fdae61",
+    5: "#d7191c",
+}
+snow_status = {2: False, 4: True}
+
+
+async def fetch_layer_data(url, session):
```
brucecrevensten, line 42: This code shouldn't be duplicated across different route files.  Pull it up to a `common.py` or something.

routes/permafrost.py
brucecrevensten: We also need to add an Explainer / About template page for this.

routes/fire.py
brucecrevensten: I made another issue to track this, so we don't have to worry about it for the moment.

brucecrevensten (approved):
I've made separate tickets/notes for things that we can address in future revisions, but this is OK for now!

PR #9, opened 2021-07-30
Modifies code so the API can be deployed to Elastic Beanstalk

brucecrevensten:
Adds requirements.txt, provides the `application` object that EB expects, adds a bit of stuff to the `.gitignore`


charparr (approved):
Looks straight-forward to me. Make it so!

PR #14, opened 2021-08-19
Adds IEM point query endpoint

brucecrevensten:


templates/iem/abstract.html
```diff
@@ -0,0 +1,17 @@
...
+    <li><a href="/iem/point">Point Query</a></li>
+    <li><a href="/iem/bbox">BBOX Query</a></li>
+    <li><a href="/iem/huc">Aggregate by HUC</a></li>
+</ul>
+<h4>/iem/point</h4>
+<p>Make a point query</p>
+<h4>/iem/bbox</h4>
+<p>Make a query with a bounding box</p>
+<h4>/iem/huc</h4>
+<p>not yeat available</p>
```
brucecrevensten, line 16: "yeat" > "yet"

kyleredilla: fixed in newer commit

PR #23, opened 2021-08-24
Add glacier, geology, physiography endpoints for Arctic-EDS

charparr:
This PR adds three endpoints for use in the development of the "Plus 5 plates for the Arctic-EDS."
Each endpoint has a `.py` route accompanied by some HTML files to aid discovery via the index.


brucecrevensten (changes requested):
Looking + working _awesome_.  I suggest:

 * Check my example additional documentation for endpoints.  Make an initial version of these for each new endpoint (don't worry about them being perfect or complete).
 * Where possible, consider adding a "code" (numeric integer etc) representing classification, etc, via a lookup table or something.  Strictly optional because I don't fully know how we'd use these, but it's a concept that's worth considering -- APIs that return only text can be a little difficult to integrate with from a client GUI perspective because you're falling back to string matching to detect classifications, etc, which can be error-prone compared to saying "it's type 1"

brucecrevensten:
Also, reminder to self, review this alongside the docs we've got to confirm a match with intent and execution.

PR #28, opened 2021-08-31
Add forest endpoint

charparr:
This PR adds a forest endpoint for the development of the "Plus 5" Arctic-EDS plate that replaces the original "Forest Types of Alaska" plate. The endpoint is fashioned in the same style as the other Arctic-EDS endpoints and references a single data layer at this time, although it uses a look-up table to retrieve several parameters based on the numeric encoding of the GeoTIFF. There may be too much vegetation information in this endpoint, but not all of it has to be exposed in the Arctic-EDS.


routes/forest.py
```diff
@@ -0,0 +1,71 @@
...
+
+forest_api = Blueprint("forest_api", __name__)
+
+wms_targets = ["ak_veg_wetland_composite"]
+wfs_targets = {}
+
+
+def package_akvegwetland(akvegwet_resp):
+    """Package forest data in dict"""
+    print(akvegwet_resp)
```
brucecrevensten, line 24: Delete debugging and/or move it into the logger if you want it to be kept!

brucecrevensten (approved):

PR #29, opened 2021-08-31
Update base templates for HTML, adds minimal styling via Bulma

brucecrevensten:
Go to each endpoint and ensure that it still works and I didn't break the demo queries.  Ensure that all pages are styled.


charparr (approved):
These look great. All queries still function as expected and all pages are styled. All links in the foot are working.

Issue #50, opened 2021-11-30
Implement CSV return for all endpoints.


charparr:
Consider making it the default, rather than JSON. This would avoid implementation in the web app.

brucecrevensten:
We won't want this for _all_ endpoints (not ones that return min/mean/max, for example, because that's trivial) but for many endpoints which return deep data that could be useful in CSV format, we'd want to implement this.

All endpoints that implement CSV output should have that documented in their relevant page.  (I missed this on the Sea Ice %Conc branch recently).

cstephen:
I took an inventory of the endpoints that do not support CSV output and it looks like we're in pretty good shape.

The output of these endpoints is too brief (summary level) for meaningful CSV output IMO:

- Elevation Point Query
- Geology Point Query
- Glaciology
- Mean Annual Precipitation
- Physiography
- Wildfire

Also, most of the "Physical and Administrative Boundary Polygons" endpoints return GeoJSON data, so CSV support is not applicable for those.

The one case I found where I think we should add CSV support is the `/places/all` endpoint. Several other endpoints link to the `/places/all` endpoint as the place where users can look up place IDs. But most users will just see an unformatted blob of JSON when they go here, so I think we should add CSV support to the `/places/all` endpoint and link to the CSV version of the URL in other parts of the documentation. I'll plan to implement this.

Issue #47, opened 2021-11-18
Allow CSV endpoints to accept a place name


brucecrevensten:
When apps request data in CSV format, the generated file should have a decent name.  Accept a `name` parameter, URL-encoded, which can be used to generate a filename.

cstephen:
This was implemented slightly differently in branch `csvs_for_iem_webapp`. In this branch, the API is using the community ID, HUC ID, or protected area ID to derive the name of the community from the API's location JSON files. This functions as input validation and ensures that CSV filenames and the location name in the CSV metadata can only be named after our list of communities. However, since single point community data is requested using just lat/lon, an additional `community` GET parameter is added to single point community requests from the web app side.

PR #62, opened 2022-01-10
Add permafrost huc query and refactor common code used across endpoints

### LINKED ISSUE #59: Minor correction to docs for permafrost endpoint
https://github.com/ua-snap/data-api/blob/1acdf4170791a20fef84350795d3753ec249ea28/templates/permafrost/point.html#L294

...should probably be `alt` as the key.

charparr:
This PR was motivated by needing to update the permafrost endpoint with the ability to query by HUC and fetch zonal mean values. During the course of that work it became clear that code would soon become duplicated across several endpoints.
In the spirit of keeping each thematic endpoint sparse, there are now a few different modules each endpoint can call upon:

`generate_requests` is used to create specific request (i.e. WCS) strings 
`generate_urls` is used to create the actual query URLs which sometimes rely on specific request strings
`validate_data` houses some minimal `nodata` handling. 

In addition to these three new modules quite a bit of functionality that was residing in individual endpoints has been moved to the existing `fetch_data` module.

All of the example endpoint queries in the clickable GUI function in my testing.

What is NOT in this PR:
 - a CSV export for permafrost and a refactoring of the CSV code if appropriate
   - see #47
   - see #50
 - a 5-model average of the GIPL data via a WCPS request
  - a revised `taspr` endpoint - I've largely left this untouched to avoid breakage. This endpoint is by far doing the most lifting at present.
  - Piping in the dimensional encodings / necessary `luts` material via `ds.attrs` or a DescribeCoverage for Rasdaman-served endpoints.
  
The above are all good candidates for future branches + PRs.
 
Closes #59


templates/glacier/point.html
```diff
@@ -23,7 +23,7 @@ <h4>Results from the query above will look like this:</h4>
     <li><strong><code>pmax</code></strong>: <i>(bool)</i>, The Pleistocene Maximum Glacial Extent did/did not cover this point with ice.</li>
     <li><strong><code>now</code></strong>:
         <ul>
-            <li><strong><code>modern</code></strong>: <i>(bool)</i>, The contempoary glacial extent does/does not cover this point with ice.</li>
+            <li><strong><code>modern</code></strong>: <i>(bool)</i>, The contemporary glacial extent does/does not cover this point with ice.</li>
```
kyleredilla, line 26: :star:

templates/permafrost/huc.html
```diff
@@ -0,0 +1,352 @@
+{% extends 'base.html' %}
```
kyleredilla, line 1: We may want to think about how much of these previews we include, probably only necessary to show some of the sample response and make it easier to scroll to descriptions below it. Fine for now.

kyleredilla (approved):
This all looks great and local tests of endpoints all work, including the new permafrost HUC endpoint. These seem like great steps towards a lean and extensible API.

kyleredilla:
Going to squash and merge because a number of the commits are doing similar things (refactoring things in similar ways across endpoints).

PR #65, opened 2022-01-14
Add top level boundary endpoint

charparr:
This PR creates a top level boundary endpoint that facilitates queries to HUCs and AK protected areas (parks, refuges, reserves, etc.) for which GeoJSONs are returned. 

Notes
 - `huc` is removed from the top-level and nested within `boundary`
 - No other endpoints besides the former top-level `huc` should be impacted

Try these routes:
http://localhost:5000/boundary/protectedarea/NPS12
http://localhost:5000/boundary/protectedarea/FWS1
http://localhost:5000/boundary/protectedarea/AKDNR9
http://localhost:5000/boundary/protectedarea/

http://localhost:5000/boundary/huc/huc8/19070506
http://localhost:5000/boundary/huc/huc8/19010208


templates/boundary/abstract.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'base.html' %}
+{% block content %}
+<h3>Boundary data</h3>
+<p>These data are intended for exposing the physical, administrative, and political polygon boundaries used across SNAP tools and data.</p>
```
kyleredilla, line 4: I think somewhere here we want to say what coordinate system the data are in. Would even be a neat feature to allow specifying the coordinate system to download in and shouldn't be that hard with a simple `to_crs()` call on the geodataframe. See forthcoming comment on PR about choosing a common default CRS.

kyleredilla (changes requested):
We should choose a default CRS to download these in.

kyleredilla:
It looks like these new boundaries are different from the HUC8 CRS, which I think is just wgs84. We should choose one, might be best to make the new boundaries come back with wgs84 too by default. Should be able to just do a quick `.to_crs(4326)` call in `luts.py`? I would be fine with defaulting to 3338 instead though. 

Also we could consider adding `?crs=<EPSG code>` modifier for the query as well??

charparr:
I suppose if you came into this API you would expect the default CRS to be 4326! Although in practice, nearly all of work is done in 3338. Whenever a zonal query occurs you can have the CRS pass from the dataset to the polygon via `.to_crs()' as you suggested.

kyleredilla:
Okay. Let's default to holding the geodataframes in memory in EPSG:3338 (so like, when the API starts and loads these shapefiles, reproject to 3338) so transformation doesn't have to happen with any of the zonal summaries. Then we transform only for the actual GeoJSON endpoints (to 4326)

charparr:
So the shape files themselves would be 4326? And then in `luts.py` they'd be converted to 3338 which would set up them up for use in all the actual data endpoints. But for the actual GeoJSON endpoints (`huc`, `protected area`) they would remain 4326?

kyleredilla:
I was thinking it doesn't matter what the shapefiles themselves are in, just do an initial reproject to 3338 on loading for those that aren't natively 3338. Then for all boundary queries, just use a `to_crs(4326)` on the polygon being requested.

charparr:
OK I think those commits implement what we discussed here!  `boundary` endpoints provide lat-lon geojsons and the data endpoints still work.

routes/boundary.py
```diff
@@ -0,0 +1,78 @@
...
+    Args:
+        huc8_id (int): HUC-8 ID
+
+    Returns:
+        GeoJSON of the HUC-8 polygon
+
+    Notes:
+        example: http://localhost:5000/boundary/huc/huc8/19070506
+    """
+    poly = huc8_gdf.to_crs(4326).loc[[huc8_id]]
```
kyleredilla, line 58: sweet! I'm gonna push a little change to swap the order of operations here from reproject/query to query reproject, it's doing a bit more than it needs as it reprojects the entire geodataframe for each query.

kyleredilla (approved):

kyleredilla:
This is looking good and is ready to merge. I've tested those GeoJSONs with some other AK shapefiles in QGIS and things look good.

PR #87, opened 2022-02-08
Add elevation endpoint

### LINKED ISSUE #40: Add elevation data (DEM) to API
We want to be able to query elevation for points, as well as providing min/max/mean for areas.  To do this, we may be able to leverage existing APIs, or, we'd need to load a local DEM and query it directly.  Target resolution at 2km since that's going to align with a lot of our existing data; 771m may be another option, depending on regridding/etc.

charparr:
This PR adds elevation data to the API via service endpoints for point and polygon (HUC + protected area) queries. The queries talk to a 3 band GeoTIFF of elevation data on GeoServer. Each band is derived from the same source data (30 m resolution ASTER GDEM) but differs in the interpolation method used to downsample from 30 m to 1 km spatial resolution. Min, max, and average interpolations were used to facilitate queries that will retain better fidelity to the actual distribution of elevation values found in the native resolution data.

Closes #40 and see that same issue for a discussion of data provenance.

This PR should also unblock https://github.com/ua-snap/iem-webapp/issues/55


templates/elevation/abstract.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'base.html' %}
+{% block content %}
+<h3>Elevation Data</h3>
+<p>These data are for use in the Northern Climate Reports tool. Currently the service endpoints query ASTER GDEM elevation data for Alaska and Western Canada. Each query fetches three values: minimum, maximum, and mean elevations derived from corresponding interpolations used to resample the source data from 30 m spatial resolution to 1 km. This technique is used to assure elevation fidelity between the original and downsampled data.</p>
```
cstephen, line 4: I think the official name of the web app is Climate Impact Reports now (despite the domain name). We probably want to change the title here.

templates/elevation/point.html
```diff
@@ -0,0 +1,43 @@
...
+<h3>Elevation Point Query</h3>
+<p>Query the ASTER Global Digital Elevation Model (DEM) at a single point specified by latitude and longitude.</p>
+<p>Example: <a href="/elevation/point/65.0628/-146.1627">/elevation/point/65.0628/-146.1627</a></p>
+<h4>Results from the query above will look like this:</h4>
+<pre>
+{
+  "res": "1 kilometer",
+  "title": "ASTER Global Digital Elevation Model",
+  "units": "meters difference from sea level",
+  "z": {
```
cstephen, line 12: The point queries return the min/avg/max data nested inside a `z` key with a `elevation_avg_` prefix for each value. The HUC and protected area queries are not nested under a `z` key and do not have an `elevation_avg_` prefix. We should make these consistent. I prefer the latter approach FWIW.

routes/elevation.py
```diff
@@ -0,0 +1,190 @@
+import asyncio
+from flask import (
+    abort,
```
cstephen, line 3: There are a handful of imports here that are not used, accordingly to Pylint:
```
elevation.py:2:0: W0611: Unused abort imported from flask (unused-import)
elevation.py:2:0: W0611: Unused Response imported from flask (unused-import)
elevation.py:2:0: W0611: Unused request imported from flask (unused-import)
elevation.py:2:0: W0611: Unused current_app imported from flask as app (unused-import)
elevation.py:15:0: W0611: Unused fetch_data imported from fetch_data (unused-import)
elevation.py:21:0: W0611: Unused project_latlon imported from validate_request (unused-import)
elevation.py:27:0: W0611: Unused nullify_nodata imported from validate_data (unused-import)
elevation.py:27:0: W0611: Unused postprocess imported from validate_data (unused-import)
```

Let's remove imports that are not used. We can add them back later if/when needed.

cstephen (changes requested):

PR #94, opened 2022-02-11
Add alfresco endpoints

### LINKED ISSUE #41: Add wildfire data to API
This would be ALFRESCO outputs to support the IEM web app.  It doesn't need to live under that endpoint, though.

kyleredilla:
This PR adds data derived from 1km ALFRESCO runs done for the IEM project to the API via service endpoints for point and polygon (HUC + protected area) queries. They pull from four coverages in Rasdaman: a historical coverage and a future coverage for each of relative flammability and relative vegetation change. Separate coverages were used for historical / future because the historical data are shared across all combinations of model / scenario. These endpoints combine the historical and future data into a single data package, for year spans / periods / eras that closely resemble what is available for the temperature and precipitation endpoints. These values are described in the "abstract" page for ALFRESCO.

There is some new general shared code created that might be useful for other endpoints to make for cleaner code, as well as some code for the new endpoints that is close to being duplicated with what is available for other endpoints (e.g. taspr), but tweaking those seems best left to another "cleanup / deduplicate" PR.

closes #41


fetch_data.py
```diff
@@ -283,3 +310,134 @@ def get_from_dict(data_dict, map_list):
...
+def parse_meta_xml_str(meta_xml_str):
+    """Parse the DescribeCoverage request to get the XML and
+    restructure to dict
+    
+    Args:
+        meta_xml_str (str): decoded text XML response from
+            DescribeCoverage WCS query
+    """
+    meta_xml = ET.ElementTree(ET.fromstring(meta_xml_str))
+    # wow xml
```
brucecrevensten, line 324: D:

kyleredilla: He I think the only option for DescribeCoverage requests is XML!! sad

templates/alfresco/abstract.html
```diff
@@ -0,0 +1,135 @@
...
+    <td>Relative Vegetation Change</td>
+    <td>These data are derived from the vegetation class data, the principal phenomena being modeled by ALFRESCO. They are computed as the average number of year-to-year transitions of any vegetation class for a pixel, across all reps for all years of the summary year span. Pixels are 1km x 1km.</td>
+  </tr>
+</tbody>
+</table>
+
+<h4>Service endpoints</h4>
+<p>Here are the existing endpoints for the above datasets.</p>
+<ul>
+    <li><a href="/alfresco/point">Point query</a>: </li>
```
brucecrevensten, line 35: Remove colons here

routes/alfresco.py
```diff
@@ -0,0 +1,478 @@
...
+    if validation == 400:
+        return render_template("400/bad_request.html"), 400
+    try:
+        huc_pkg = run_aggregate_var_polygon(var_ep, huc8_gdf, huc_id)
+    except:
+        return render_template("422/invalid_huc.html"), 422
+
+    if request.args.get("format") == "csv":
+        varname = var_ep_lu[var_ep]["varname"]
+        csv_data = create_csv(huc_pkg, alf_fieldnames, varname=varname, stat="mean")
```
brucecrevensten, line 446: Here and in the point & protected area, the columns headers don't match the data (opened in Numbers on MacOS), but the CSV agrees.

Screen Shot 2022-02-15 at 11 09 25 AM

brucecrevensten (changes requested):

fetch_data.py
```diff
@@ -34,6 +37,21 @@ async def fetch_wcs_point_data(x, y, cov_id, var_coord=None):
     return point_data
 
 
+async def fetch_wcps_point_data(x, y, cov_id, axis_name, axis_coords):
```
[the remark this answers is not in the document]
kyleredilla: removed!

templates/alfresco/abstract.html
kyleredilla: done

routes/alfresco.py
kyleredilla: fixed

fetch_data.py
```diff
@@ -283,3 +301,141 @@ def get_from_dict(data_dict, map_list):
...
+
+    Returns:
+        CSV Response
+    """
+    response = Response(
+        csv_data,
+        mimetype="text/csv",
+        headers={
+            "Content-Type": 'text/csv; name="climate.csv"',
+            "Content-Disposition": 'attachment; filename="climate.csv"',
```
charparr, line 438: I don't know that it is necessary now but we  may want an optional argument for the `filename` and `name` key-value pairs in the `headers` dictionary.

generate_requests.py
```diff
@@ -64,3 +69,66 @@ def generate_netcdf_wcs_getcov_str(bbox_bounds, cov_id, var_coord=None):
...
+    Args:
+        bbox_bounds (tuple): 4-tuple of bounding polygon extent (xmin, ymin, xmax, ymax)
+        generate_average_wcps_str_kwargs (dict): Args to pass on to generate_average_wcps_str_kwargs
+
+    Returns:
+        netcdf_avg_wcps_str (str): WCPS GetCoverage Request to append to a query URL
+    """
+    (x1, y1, x2, y2) = bbox_bounds
+    x = f"{x1}:{x2}"
+    y = f"{y1}:{y2}"
```
charparr, line 130: Some duplication with the `get_wcs_xy_str_from_bbox_bounds` but you'd have to arrange it such that this function uses a `Polygon` as the input vs. the current `bbox_bounds`. Each endpoint that generates netCDF WCPS averages would have to get changed...so it might be better left for a separate PR.

charparr (changes requested):
Most of what I've noted here should be addressed later on - I think just the two actual suggestions (very minor doc fixes) should be changed before merging. I'm able to run a bunch of queries on this endpoint with success!

generate_requests.py
kyleredilla: Ohhh good catch. I made a note of this one in issue #97  - it's high time we had another de-duplication / consistency PR.

routes/alfresco.py
```diff
@@ -0,0 +1,495 @@
...
+
+    data_arr = []
+    for coords in iter_coords:
+        sel_di = {dimname: int(coord) for dimname, coord in zip(dimnames, coords)}
+        data_arr.append(da.sel(sel_di).values)
+    data_arr = np.array(data_arr)
+
+    # need to transpose the 2D spatial slices if X is the "rows" dimension
+    if all_dims.index("X") < all_dims.index("Y"):
+        data_arr = data_arr.transpose(0, 2, 1)
```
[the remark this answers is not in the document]
kyleredilla: Yep good call also commented on #97

charparr (approved):

routes/alfresco.py
```diff
@@ -0,0 +1,478 @@
...
+
+
+@routes.route("/alfresco/flammability/protectedarea/")
+@routes.route("/alfresco/veg_change/protectedarea/")
+@routes.route("/alfresco/protectedarea/")
+def alfresco_about_protectedarea():
+    return render_template("alfresco/protectedarea.html")
+
+
+@routes.route("/alfresco/<var_ep>/point/<lat>/<lon>")
```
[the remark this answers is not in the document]
BobTorgerson: This [line](https://github.com/ua-snap/data-api/blob/c047b7f20fde42229a5e484d0753c1cd766294fe/routes/recache.py#L122) grabs all of the routes in the application. Then the rest of the processing checks for the string "point" and will do all of the transforming of the URL to get every individual endpoint. 

Nothing needs to be changed for the recache function to pick up this new endpoint.

routes/alfresco.py
```diff
@@ -0,0 +1,478 @@
...
+    )
+
+    if request.args.get("format") == "csv":
+        csv_data = create_csv(point_pkg, alf_fieldnames, varname=varname, stat="mean")
+        return return_csv(csv_data)
+
+    return postprocess(point_pkg, "alfresco")
+
+
+@routes.route("/alfresco/<var_ep>/huc/<huc_id>")
```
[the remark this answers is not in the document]
BobTorgerson: This also will be grabbed by the current recache code. Nothing needs to be done for this to be picked up.

PR #124, opened 2022-03-02
Additions to the README specific to the /update/ endpoint.

### LINKED ISSUE #123: Add documentation of "/update" endpoint
When we start up the API locally now, we should always know to run /update as a first step (and perhaps we want to automate this!).

Also note that the returned JSON from that endpoint isn't valid JSON so it kinda makes some tools blow up.

BobTorgerson:
Fixes #123

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,16 @@ Example Geology Query
 Example Physiography Query
  - http://localhost:5000/physiography/64.606/-147.345
 
+## Updating the JSON data on the API
+
+We have an endpoint specifically meant to pull down data
+from our geospatial truth repository. This will automatically
+generate GeoJSON files based on those CSVs and shapefiles.
+
+To get the latest data, go to:
+
+ - http://localhost:5000/update/
+
 ## Updating Production API + Varnish Cache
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,6 +38,16 @@ Example Geology Query
 Example Physiography Query
  - http://localhost:5000/physiography/64.606/-147.345
 
+## Updating the JSON data on the API
+
+We have an endpoint specifically meant to pull down data
```
brucecrevensten, line 43: This is good; I suspect that there's been a bit of a change in other concurrent PRs to make this happen automatically (unsure); also it'd be good to indicate here that this endpoint is called against the production API via an Action when code is merged into the geospatial vector veracity repo.

brucecrevensten (commented):
I'll hold on merging this until the other open work is merged in case it impacts this.

README.md
brucecrevensten: x-ref #145, #146

brucecrevensten:
@BobTorgerson can you check this again when you can, make sure that all the documentation in this PR is accurate after all the other work that was done, then merge?

PR #131, opened 2022-03-07
Use dashes instead of underscores for ALFRESCO polygonal era keys

### LINKED ISSUE #130: Use dashes, not underscores, for polygonal ALFRESCO era keys

cstephen:
Closes #130.

Point queries return decadal and 30-year era ALFRESCO data using JSON keys with dashes (e.g., "2040-2069"). Polygonal queries currently return decadal data with dashes and era data with underscores (e.g., "2040_2069"). This updates the polygonal queries to use dashes to be consistent with the other keys from the ALFRESCO endpoint.


kyleredilla (approved):
Oops! Looks good / tests good.

PR #132, opened 2022-03-07
Generalize shapefile ingest and return precision controlled GeoJSON

### LINKED ISSUE #104: Update JSON dict in luts.py for other boundary endpoints
https://github.com/ua-snap/data-api/blob/8a46d6e88e4ca0f7d7f418acb69f646c23619d8b/luts.py#L74

### LINKED ISSUE #111: Remove data/csvs/communities
Remove this file from version control because it's expected to be dynamically fetched as needed from another repo.

### LINKED ISSUE #121: Clamp precision for returned GeoJSON to 4 or 5 digits of precision
4 is probably enough, 5 may be better; the goal is to reduce the amount of floating point noise that we're shipping over the network to reduce the payload size for any requests to the `/boundary/xxx` to get GeoJSON for a polygon.

charparr:
This PR generalizes the existing logic for fetching shapefiles from GitHub and generating appropriate json structures via adictionary mapping in `luts.py`. The PR also clamps GeoJSON precision across all boundary endpoints and moves to ignore data jsons, csvs, and shapefiles in this repo. I successfully ran the `update` endpoint prior to this commit. 

Closes #121 
Closes #104 
Closes #111


charparr:
Not bad if that is the only conflict ^ and probably a good justification for only tracking things in the repo and forcing ourselves to run `update` over here whenever we make changes to the backend geospatial data.

PR #133, opened 2022-03-07
Preserve keys for null values if any sibling contains data in pruned response

### LINKED ISSUE #120: Preserve permafrost "alt" keys when there is no data
Repeating ALT nodata values like 0.07 have been converted to actual nodata values before the Rasdaman ingest. Xref #85. These "alt" keys & values are now completely pruned off the response JSON because they are nested several layers deep and our response pruning algorithm is set to prune permafrost nodata at a depth of 2. We need to figure out a way to make our pruning algorithm a little more intelligent.

cstephen:
Closes #120.

This PR changes the way that our response pruning algorithm works. In the current main branch, the pruning function allows us to set the minimum pruning depth so that certain keys are preserved even if they have a null value. Keeping certain keys intact makes it easier to handle the data on the client-side without inspecting the structure of the entire JSON object.

For example, we want the "gipl" key (and "jorg" and "obupfx" keys) to be preserved in the following response even though they have null values:

https://earthmaps.io/permafrost/point/52.9025/172.909444

The downside to this approach is that the pruning algorithm is too aggressive at higher nesting depths. For example, we want the "alt" key to be preserved in this response even though the alt values are null:

https://earthmaps.io/permafrost/point/61.0628/-147.1627

Since there is no obvious way to set two separate pruning depths simultaneously, a different approach is needed. Fortunately, a pattern is emerging here: We want to preserve JSON keys for null values if any of the sibling keys have non-null data. This PR implements this functionality and removes the unnecessary minimum-prune-depth feature to simplify the code considerably.

Testing:

The following URL should have keys for gipl, jorg, and obupfx even though they have null values:
https://earthmaps.io/permafrost/point/52.9025/172.909444
http://localhost:5000/permafrost/point/52.9025/172.909444

The following URL should have gipl "alt" keys even though they are all set to null:
https://earthmaps.io/permafrost/point/59.24/-135.51
http://localhost:5000/permafrost/point/59.24/-135.51

All other API URLs should continue to work as before!

### DIFF
--- a/routes/permafrost.py
+++ b/routes/permafrost.py
@@ -258,7 +258,7 @@ def run_point_fetch_all_permafrost(lat, lon):
         "obupfx": package_obu_vector(gs_results[2]),
     }
 
-    return postprocess(data, "permafrost", titles, 2)
+    return postprocess(data, "permafrost", titles)
 
 
 @routes.route("/permafrost/huc/<huc_id>")
--- a/validate_data.py
+++ b/validate_data.py
@@ -46,7 +46,7 @@ def nullify_nodata(data, endpoint):
     return nullified
 
 
-def prune_nodata_dict(data, prune_depth, depth):
+def prune_nodata_dict(data):
     """Traverse dict recursively and prune empty or None branches.
 
     Args:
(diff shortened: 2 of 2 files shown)


kyleredilla (approved):
These changes look good and greatly simplify the code while achieving a more useful algorithm for the API! I've tested the new endpoints and compared with the existing ones, and the new code appears to work as expected.

PR #134, opened 2022-03-08
Convert elevation nodata to null in JSON response

### LINKED ISSUE #125: Nullify / prune elevation nodata values and serve HTTP 404 errors where appropriate
We need to nullify & prune nodata values for the new elevation endpoint similar to what is done for the other endpoints.

Here's an elevation data URL with nodata values for example:
https://earthmaps.io/elevation/point/68.55/-174.88

cstephen:
Closes #125.

This PR is dependent on #133, so I'm setting this PR as a draft until #133 is merged.

To test, compare this URL between production and the `elevation_nodata` branch running locally:

https://earthmaps.io/elevation/point/68.55/-174.88
http://localhost:5000/elevation/point/68.55/-174.88

Instead of this:

```
{
  "max": -9999,
  "mean": -9999,
  "min": -9999,
  "res": "1 kilometer",
  "title": "ASTER Global Digital Elevation Model",
  "units": "meters difference from sea level"
}
```

You should see this:

```
{
  "max": null,
  "mean": null,
  "min": null,
  "res": "1 kilometer",
  "title": "ASTER Global Digital Elevation Model",
  "units": "meters difference from sea level"
}
```

### DIFF
--- a/routes/elevation.py
+++ b/routes/elevation.py
@@ -15,7 +15,7 @@ from validate_request import (
     validate_huc,
     validate_akpa,
 )
-from validate_data import get_poly_3338_bbox
+from validate_data import get_poly_3338_bbox, postprocess
 from config import GS_BASE_URL, WEST_BBOX, EAST_BBOX
 from luts import huc_gdf, akpa_gdf
 from . import routes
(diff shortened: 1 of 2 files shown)


kyleredilla (approved):
These changes look good and appears to have the desired effect.

PR #139, opened 2022-03-16
Permafrost CSV exports implemented, metadata added to CSVs, and CSV 404 checks working.

### LINKED ISSUE #47: Allow CSV endpoints to accept a place name

cstephen:
Closes #47.

This PR depends on the `csv_improvements` branch of the iem-webapp.

To test, run the data-api like this:

```
git checkout csvs_for_iem_webapp
export WEB_APP_URL=http://localhost:3000/
pipenv run flask run
```

And run the iem-webapp like this:

```
git checkout csv_improvements
export SNAP_API_URL=http://localhost:5000
npm run dev
```

Visit the iem-web app at http://localhost:3000 and load a report for each of the following:
- A lat/lon point (not a pre-defined community)
- A community
- A HUC
- A protected area
- A fire management unit
- A corporation
- A climate division
- An ethnolinguistic region

For each of the types of report above, verify that there is a CSV download button for each of the following types of data (if they are displayed on the report page):

- Temperature
- Precipitation
- Permafrost
- Flammability
- Vegetation change

Load each of the CSV downloads into a spreadsheet application and verify that the metadata, columns, and values make sense. Also verify that the filename of each CSV file is named after the corresponding place name.

I did some research and there appears to be no standard way to add metadata/comments to a CSV file. Currently, metadata is added as lines starting with "#" at the beginning of the CSV.


fetch_data.py
```diff
@@ -263,7 +265,11 @@ def summarize_within_poly(ds, poly, dim_encodings, varname="Gray", roundkey="Gra
 
 def geotiff_zonal_stats(poly, arr, transform, stat_list):
     poly_mask_arr = zonal_stats(
-        poly, arr, affine=transform, nodata=np.nan, stats=stat_list,
+        poly,
```
kyleredilla, line 268: Minor thing but what is causing changes like this? Is this `black`? I feel like I've seen formatting stuff like this going back and forth in various parts. It might just be line length. I think my `black` defaults to line lengths of 88, so this gets set back to the way it was originally if I were to run `black` on it.

fetch_data.py
```diff
@@ -419,25 +439,109 @@ def create_csv(packaged_data, fieldnames, package_coords, fill_di=None):
...
         row_di["value"] = get_from_dict(packaged_data, coords)
-        writer.writerow(row_di)
-    return output.getvalue()
+        rows.append(row_di)
+    return rows
 
 
-def return_csv(csv_data):
-    """Return the CSV data as a download
+def write_csv(csv_dicts, fieldnames, filename, metadata=None):
```
kyleredilla, line 446: Nice, this all looks good and makes sense to divy it up among the new functions this way.

fetch_data.py
```diff
@@ -419,25 +439,109 @@ def create_csv(packaged_data, fieldnames, package_coords, fill_di=None):
...
-            "Content-Type": 'text/csv; name="climate.csv"',
-            "Content-Disposition": 'attachment; filename="climate.csv"',
+            "Content-Type": 'text/csv; name="' + filename + '"',
+            "Content-Disposition": 'attachment; filename="' + filename + '"',
         },
     )
     return response
+
+
+def place_name(place_type, place_id):
```
kyleredilla, line 475: Sweet. It feels like the `validate_data.py` module might be a better fit for this? Fine to me as is though.

routes/permafrost.py
```diff
@@ -296,6 +361,26 @@ def run_huc_fetch_all_permafrost(huc_id):
     magt_huc_pkg = package_gipl_polygon(magt_poly_sum_di)
     alt_huc_pkg = package_gipl_polygon(alt_poly_sum_di)
     combined_pkg = combine_gipl_poly_var_pkgs(magt_huc_pkg, alt_huc_pkg)
+
```
kyleredilla, line 364: Maybe use a `create_csv` function like what was used in `routes/alfresco.py` to do some deduplication? At least within this module and any others for consistency, but could potentially be imported `create_csv` function just can't tell from a quick look how generalizable it is.

fetch_data.py
cstephen: I think you're right, `validate_data.py` is a better place for this. I'll get this fixed!

cstephen: Fixed in 6e2b6c3b4f9b8c5deab7212f69190a3b592f75ba.

fetch_data.py
cstephen: Yep, definitely a Black thing. This line of code was 69 characters long before the Black reformatting, so it doesn't seem to be a line length issue. I think Black maybe just doesn't like function parameters listed on their own line like this. I made sure I'm running the latest version of Black (22.1.0) with default settings, so I'll keep this change as-is for now. Not sure what's going on or why our version are acting differently.

routes/permafrost.py
cstephen: I wasn't able to get this done before the PR was merged, but I've created issue #147.

fetch_data.py
```diff
@@ -419,25 +438,72 @@ def create_csv(packaged_data, fieldnames, package_coords, fill_di=None):
...
+    """
+    metadata = "# Location: "
+    if place_name is None:
+        metadata += lat + " " + lon + "\n"
+    elif place_type == "communities":
+        metadata += place_name + "\n"
+    else:
+        metadata += place_name + " (" + place_type_labels[place_type] + ")\n"
+
+    report_url = WEB_APP_URL + "report/"
```
brucecrevensten, line 500: OK I just realized now that we *probably* want to move this code into a string that is provided to the API, this way the API doesn't need any know anything about the URL structure of the client app.  (This isn't wrong, it's just nice to keep them as separate as possible).

routes/permafrost.py
```diff
@@ -236,4 +253,63 @@ def run_point_fetch_all_permafrost(lat, lon):
...
+        place_name, place_type = place_name_and_type(place_id)
+
+        metadata = csv_metadata(place_name, place_id, place_type, lat, lon)
+        metadata += "# alt is the active layer thickness in meters\n"
+        metadata += "# magt is the mean annual ground temperature in degrees Celsius\n"
+        metadata += "# ice is the estimated ground ice volume\n"
+        metadata += "# pfx is the permafrost extent\n"
+        metadata += "# 2025 represents 2011 - 2040\n"
+        metadata += "# 2050 represents 2036 - 2065\n"
+        metadata += "# 2075 represents 2061 – 2090\n"
```
brucecrevensten, line 301: Need to be changed to regular dashes just to match, but I made another ticket for that.

routes/taspr.py
```diff
@@ -442,21 +484,33 @@ def create_csv(packaged_data):
...
+        lat: latitude unless an area
+        lon: longitude unless an area
 
     Returns:
         CSV Response
     """
+
+    place_name, place_type = place_name_and_type(place_id)
+
+    if place_name is not None:
```
brucecrevensten, line 503: Maaaaaaybe we make a ticket to pull all of this related logic into a single place, since I see similar fragments throughout this PR.  Not sure.

brucecrevensten (approved):

Issue #154, opened 2022-04-08
Improve documentation + add example output for new veg_type ALFRESCO endpoint


cstephen:
A little bit of documentation regarding the new `veg_type` endpoint has been added already, but it differs enough from the other two ALFRESCO endpoints (`flammability` and `veg_change`) that we'll need a more careful revamp of the documentation for ALFRESCO endpoints.

PR #155, opened 2022-04-08
Add new vegetation type (aka veg_type) endpoint that returns each vegetation type as a percentage

cstephen:
To test, use the new `veg_type` endpoint for any lat/lon or area query that works for the existing ALFRESCO endpoints (`flammability` and `veg_change`):

http://localhost:5000/alfresco/veg_type/

This new endpoint can use some better documentation and example output. This has been captured in #154.

### DIFF
--- a/routes/alfresco.py
+++ b/routes/alfresco.py
@@ -39,29 +39,31 @@ alfresco_api = Blueprint("alfresco_api", __name__)
 summary_era_coords = [(3, 5), (6, 8)]
 
 # create encodings for coverages (currently all coverages share encodings)
-future_dim_encodings = asyncio.run(get_dim_encodings("relative_flammability_future"))
-historical_dim_encodings = asyncio.run(
+flammability_future_dim_encodings = asyncio.run(get_dim_encodings("relative_flammability_future"))
+flammability_historical_dim_encodings = asyncio.run(
     get_dim_encodings("relative_flammability_historical")
 )
+veg_type_dim_encodings = asyncio.run(get_dim_encodings("alfresco_vegetation_type_percentage"))
 
 var_ep_lu = {
     "flammability": {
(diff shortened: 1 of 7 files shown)


routes/alfresco.py
```diff
@@ -325,91 +396,150 @@ def run_aggregate_var_polygon(var_ep, poly_gdf, poly_id):
...
-        {"variable": varname, "stat": "mean"},
+        data_pkg, fieldnames, {"variable": varname, "stat": stat},
     )
 
     place_name, place_type = place_name_and_type(place_id)
 
     metadata = csv_metadata(place_name, place_id, place_type, lat, lon)
 
-    if var_ep == "flammability":
+    if var_ep in "flammability":
```
kyleredilla, line 498: Is this right? Or should it be equivalence:
```suggestion
    if var_ep == "flammability":
```

routes/alfresco.py
```diff
@@ -538,21 +671,31 @@ def run_fetch_alf_local_data(var_ep, lat, lon):
...
+
     huc12_pkg["huc_id"] = huc_id
     huc12_pkg["boundary_url"] = f"https://earthmaps.io/boundary/area/{huc_id}"
 
     return huc12_pkg
 
 
 @routes.route("/alfresco/<var_ep>/area/<var_id>")
-def run_fetch_alf_area_data(var_ep, var_id):
+def run_fetch_alf_area_data(var_ep, var_id, ignore_csv=False):
```
kyleredilla, line 690: Added this to handle HUC12 local CSVs. If you think of a strategy you like more, please go for it!

kyleredilla (approved):
I've tested this new endpoint and it works well! The changes all look good, I approve this for merging as is but left some optional comments.

PR #161, opened 2022-04-08
Update `boundary` documentation and condense `elevation` polygon queries

### LINKED ISSUE #156: Add HUC12s to boundary endpoint
Need to add ability to query `/boundary` endpoint for HUC12 & get GeoJSON back.

Also double-check the inline documentation in the boundary.py route file, it looks outdated where it gives an example.

### LINKED ISSUE #159: Update elevation templates + routes
Documentation templates and routes should reflect the change to accept general "area" queries to get summarizes by HUC, fire mgmt, protected area, etc.

charparr:
This small PR updates the `boundary` documentation to reflect the existence of HUC-12 geoJSON queries that are accessed via `area/twelvedigitnumber` where the latter is a valid AK HUC-12 integer code.  This PR also updates the `elevation` route and documentation to match the current, and more general, `area` routing pattern (vs. using separate endpoints for huc, protected area, etc.). 

For example, http://localhost:5000/elevation/area/19010208 was previously http://localhost:5000/elevation/huc/19010208

Closes #159 
Closes #156


kyleredilla (approved):
These changes look good! I've tested with a variety of different area types and things seem to work.

PR #163, opened 2022-04-14
Use new flammability coverage with precomputed eras, streamline ALFRESCO output, and fix polygon mask shape mismatches

### LINKED ISSUE #153: Decide whether to keep both vegetation change & vegetation percent endpoints, or just one
We can't decide this until we get feedback regarding the iem-webapp vegetation type percentage charts. If we do end up keeping the vegetation type percentage charts, the web app will no longer be using the vegetation change (veg_change) API endpoint, only the percentage type percent endpoint (veg_type). We probably don't want to maintain the veg_change endpoint unless it will be used for something else in the future.

### LINKED ISSUE #154: Improve documentation + add example output for new veg_type ALFRESCO endpoint

### LINKED ISSUE #160: Condense and clean up the veg_type endpoint output
Currently the new `veg_type` endpoint returns data like this:

```
"barren_lichen_moss": {
  "vt": 0
},
"black_spruce": {
  "vt": 0
},
"deciduous_forest": {
  "vt": 50.92
},
"graminoid_tundra": {
  "vt": 0
},
"not_modeled": {
  "vt": 0
},
"shrub_tundra": {
  "vt": 0
},
"temperate_rainforest": {
  "vt": 0
},
"wetland_tundra": {
  "vt": 0
},
"white_spruce": {
  "vt": 49.08
}
```

The `vt` keys for each percentage value stands for "vegetation type", which doesn't really make sense here. It ended up this way because of overaggressive code re-use.

We can make this much better by returning output like this instead:

```
{
  "barren_lichen_moss": 0,
  "black_spruce": 0,
  "deciduous_forest": 50.92,
  "graminoid_tundra": 0,
  "not_modeled": 0,
  "shrub_tundra": 0,
  "temperate_rainforest": 0,
  "wetland_tundra": 0,
  "white_spruce": 49.08
}
```

This change will need to be coordinated with a corresponding change to the iem-webapp code.

### LINKED ISSUE #162: Fix issues with polygon mask array shape not matching data array shape
This affects any endpoint that averages data over an area (including flammability, vegetation type, temperature, precipitation). If you load any area query, more often than not the polygon mask array for area will have a slightly difference shape vs. the original data. The shape mismatch causes server errors when this line of code is called:

```
data_arr_mask = np.broadcast_to(poly_mask_arr.mask, data_arr.shape)
```

If you commented out the try/except block wrapped around the line above, errors like this will show up in the console output:

```
ValueError: operands could not be broadcast together with remapped shapes [original->remapped]: (128,188)  and requested shape (140,127,187)
```

Our current workaround for this is to crop the polygon mask array to fork it into the correct shape. Can we solve this by adding a bit of buffer/padding around the polygon bounding box when the data is requested from Rasdaman?

cstephen:
Closes #153.
Closes #154.
Closes #160.
Closes #162.

This PR does several things:

- Fixes all area endpoints, which broke mysteriously. After the recent Rasdaman upgrade on Zeus (and earlier on Apollo), the shape of polygon masks suddenly did not match the shape of the original data arrays. To solve this, we are now cropping polygon masks to the correct shape.
- Switches to using the new `alfresco_relative_flammability_30yr` Rasdaman coverage that has pre-computed 30-year projected eras + two distinct historical eras (1950-1979, 1980-2008).
- Removes code that handled dynamic aggregation of ALFRESCO data into 30-year eras since this work has now been precomputed before the Rasdaman ingest.
- Removes the old `veg_change` (vegetation transitions) endpoint that has been superseded by the new `veg_type` (vegetation percentages) endpoint.
- Removes documentation regarding the old `veg_change` endpoint and adds more documentation for the new `veg_type` endpoint.
- Removes the "rf" (relative flammability) and "vt" (vegetation type) variable names from ALFRESCO output that were not quite correct, and no longer served a purpose. This flattens & streamlines the output for both types of ALFRESCO data.
- Parses the slightly different XML metadata structure introduced in part by the new version of Rasdaman, I think.

### DIFF
--- a/routes/alfresco.py
+++ b/routes/alfresco.py
@@ -37,36 +37,24 @@ from . import routes
 alfresco_api = Blueprint("alfresco_api", __name__)
 
 
-# These are the encoded coordinate values for the "era" axis of the
-# rasdaman coverages being queried here. Each tuple contains the first
-# and last coordinate for creating WCPS query to average over them.
-# E.g., (3,5) computes average across 2040-2049, 2050-2059, and 2060-2069.
-summary_era_coords = [(3, 5), (6, 8)]
-
 # create encodings for coverages (currently all coverages share encodings)
-flammability_future_dim_encodings = asyncio.run(
-    get_dim_encodings("relative_flammability_future")
+flammability_dim_encodings = asyncio.run(
+    get_dim_encodings("alfresco_relative_flammability_30yr")
 )
-flammability_historical_dim_encodings = asyncio.run(
-    get_dim_encodings("relative_flammability_historical")
-)
-
-veg_change_future_dim_encodings = flammability_future_dim_encodings
-veg_change_historical_dim_encodings = flammability_historical_dim_encodings
 
 veg_type_dim_encodings = asyncio.run(
     get_dim_encodings("alfresco_vegetation_type_percentage")
 )
 
 var_ep_lu = {
-    "flammability": {"cov_id_str": "relative_flammability", "varname": "rf",},
-    "veg_change": {"cov_id_str": "relative_vegetation_change", "varname": "rvc",},
-    "veg_type": {"cov_id_str": "alfresco_vegetation_type_percentage", "varname": "vt",},
+    "flammability": {"cov_id_str": "alfresco_relative_flammability_30yr"},
+    "veg_type": {
+        "cov_id_str": "alfresco_vegetation_type_percentage",
+    },
 }
 
 var_label_lu = {
     "flammability": "Flammability",
-    "veg_change": "Vegetation Change",
     "veg_type": "Vegetation Type",
 }
 
@@ -75,17 +63,15 @@ flammability_fieldnames = [
     "date_range",
     "model",
     "scenario",
-    "variable",
     "stat",
     "value",
 ]
-veg_change_fieldnames = flammability_fieldnames
+
 veg_type_fieldnames = [
     "date_range",
     "model",
     "scenario",
     "veg_type",
-    "variable",
     "stat",
     "value",
 ]
@@ -93,7 +79,7 @@ veg_type_fieldnames = [
 
 async def fetch_alf_point_data(x, y, cov_id_str):
     """Make the async request for the data at the specified point for
-    a specific varname.
+    a specific coverage
 
     Args:
         x (float): lower x-coordinate bound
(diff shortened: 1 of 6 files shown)


templates/alfresco/abstract.html
```diff
@@ -1,141 +1,263 @@
...
-    "5modelAvg": {
+  ...
+}
+</pre>
+
+<h4>Dimensions:</h4>
+<ul>
+  <li>
+    <strong><code>era</code></strong
+    >: Sumary era:
```
charparr, line 94:
```suggestion
    >: Summary era:
```

templates/alfresco/abstract.html
```diff
@@ -1,141 +1,263 @@
...
-            <li>CRU_historical</li>
-        </ul>
-</ul>
-<strong>Variables:</strong>
-<ul>
-    <li><strong><code>rf</code></strong>: <i>(float)</i>, relative flammability.</li>
-    <li><strong><code>rvc</code></strong>: <i>(float)</i>, relative vegetation change.</li>
+  <li>
+    <strong><code>era</code></strong
+    >: Sumary era:
```
charparr, line 208:
```suggestion
    >: Summary era:
```

charparr (changes requested):
Great PR! The endpoints are all functioning as I'd expect and the documentation is very clear. See my suggestions to nab a couple typos and a maybe now-deprecated comment and this is ready to merge.  🔥 👨‍💻

routes/alfresco.py
```diff
@@ -396,47 +318,22 @@ def run_aggregate_var_polygon(var_ep, poly_gdf, poly_id):
...
-        aggr_results = remove_invalid_dim_combos(aggr_results)
+    for era, summaries in ar5_results.items():
+        aggr_results[era] = summaries
+    aggr_results = remove_invalid_dim_combos(var_ep, aggr_results)
 
     return aggr_results
 
 
-def remove_invalid_dim_combos(results):
+def remove_invalid_dim_combos(var_ep, results):
```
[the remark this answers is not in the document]
cstephen: I like this idea a lot! This will help keep all metadata pertaining to the coverage in Rasdaman and make the API code more coverage agnostic and easier to generalize and recycle in the future. Good idea for future iteration.

routes/alfresco.py
```diff
@@ -37,36 +37,24 @@
 alfresco_api = Blueprint("alfresco_api", __name__)
 
 
-# These are the encoded coordinate values for the "era" axis of the
-# rasdaman coverages being queried here. Each tuple contains the first
-# and last coordinate for creating WCPS query to average over them.
-# E.g., (3,5) computes average across 2040-2049, 2050-2059, and 2060-2069.
-summary_era_coords = [(3, 5), (6, 8)]
-
 # create encodings for coverages (currently all coverages share encodings)
```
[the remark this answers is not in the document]
cstephen: Fixed in 0496ce0a4c8459239911af85d431d4e3f0e10567.

templates/alfresco/abstract.html
cstephen: Fixed in 0496ce0a4c8459239911af85d431d4e3f0e10567.

charparr (approved):

PR #167, opened 2022-04-22
Fix server error and return valid response if point query returns no nearby HUCs

cstephen:
Looks like this problem started occurring when the Aleutian Island HUCs were removed in https://github.com/ua-snap/geospatial-vector-veracity/pull/86.

To replicate, view a point query from a Western Aleutian Island while running the data-api from the `main` branch. For example:

http://localhost:5000/places/search/51.77/-176.67

The `no_huc_fix` branch will return a valid response, which I have tested against the web app.

Xref: https://github.com/ua-snap/iem-webapp/issues/317.


charparr (approved):
I'm able to replicate the functioning query and initializing with an empty list is a good pattern a follow when we are searching by point locations.

PR #179, opened 2022-05-17
Arctic EDS January & July Min Mean Max Temperature Endpoint

BobTorgerson:
This PR adds the ability to generate min, mean, and max values from historical and projected scenarios. 

There are 5 expected outputs from this PR:

**Historical Min-Mean-Max**
- http://localhost:5000/mmm/jan/historical/65.0628/-146.1627

**Projected Min-Mean-Max** 
- http://localhost:5000/mmm/july/projected/65.0628/-146.1627

**Historical & Projected Min-Mean-Max**
- http://localhost:5000/mmm/jan/hp/65.0628/-146.1627

**JSON dump of all available min-mean-max data for a given point**
- http://localhost:5000/mmm/july/all/65.0628/-146.1627

**CSV download of all available min-mean-max data for a given point**
- http://localhost:5000/mmm/jan/65.0628/-146.1627?format=csv

Try several places and check that the CSV output looks good!

### DIFF
--- a/generate_requests.py
+++ b/generate_requests.py
@@ -20,6 +20,44 @@ def get_wcs_xy_str_from_bbox_bounds(poly):
     return xy
 
 
+def generate_mmm_wcs_getcov_str(x, y, cov_id, model, scenario, encoding="json"):
+    """Generate a WCS GetCoverage request for fetching a
+    subset of a coverage over X and Y axes.
+
+    Args:
+        x (float or str): x-coordinate for point query (float), or string
+        composed as "x1,x2" for bbox query, where x1 and x2 are
+        lower and upper bounds of bbox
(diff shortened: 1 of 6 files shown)


cstephen (changes requested):
This is looking and working great, generally! All of the URLs in the PR description work except for this one:

http://localhost:5000/mmm/jan/65.0628/-146.1627?format=csv

But it looks like that is just because of a typo in the URL. It works if I add `/all`:

http://localhost:5000/mmm/jan/all/65.0628/-146.1627?format=csv

The only other thing I noticed is that this new endpoint is not detecting -9999 nodata values and handling them like the other endpoints do. For example, I would expect this URL to return our custom 404 page because the response is entirely -9999 values:

http://localhost:5000/mmm/july/projected/65.0628/-139.1627

PR #180, opened 2022-05-20
Adds precipitation min-mean-max endpoint and annual_precip_totals coverage CSV download

BobTorgerson:
This PR adds a new endpoint for generating the min-mean-max for precipitation coming from the annual_precip_totals coverage. 

This adds the following new endpoint:
http://localhost:5000/mmm/precipitation

In addition, the previous /mmm endpoint has been updated so that it is able to be used for temperature and precipitation. The previous endpoints for temperature are now as follows:
http://localhost:5000/mmm/temperature

The documentation found at both of those URLs above have been updated with the correct URLs for testing.


templates/mmm/abstract.html
```diff
@@ -1,10 +1,11 @@
 {% extends 'base.html' %}
 {% block content %}
-<h3>Historical and Projected Temperature Minimum, Mean, and Maximum</h3>
-<p>The endpoint here provides access to the temperature at surface in degrees Celsius for both historical and projected model data over the state of Alaska.</p>
+<h3>Historical and Projected Temperature + Precipitation Minimum, Mean, and Maximum</h3>
+<p>The endpoint here provides access to the temperature at surface in degrees Celsius and totail annual precipitation in mm for both historical and projected model data over the state of Alaska.</p>
```
cstephen, line 4: Small typo here in the word "totail".

cstephen (changes requested):
I tested the precipitation and relocated temperature endpoints, tried different coordinates, made sure "no data" responses are caught and turned into 404s, and read the documentation. Everything looks good except for the tiny typo I found (see code comments).

templates/mmm/abstract.html
BobTorgerson: Good catch Craig!

PR #194, opened 2022-06-02
Adds /eds/temperature and /eds/precipitation requests for consolidated ArcticEDS plate requests.

BobTorgerson:
This PR combines the requests for the ArcticEDS client side into a single function to have the client able to request one endpoint and receive all of the data required.

### DIFF
--- a/routes/taspr.py
+++ b/routes/taspr.py
@@ -1125,25 +1125,56 @@ def get_temperature_plate(lat, lon):
             example request: http://localhost:5000/eds/temperature/65.0628/-146.1627
     """
     temp_plate = dict()
+
+    ### HISTORICAL ###
     temp_plate['historical'] = dict()
-    temp_plate['historical']['all'] = mmm_point_data_endpoint("temperature", "historical", lat, lon)
-    temp_plate['historical']['jan'] = mmm_point_data_endpoint("temperature", "historical", lat, lon, "jan")
-    temp_plate['historical']['july'] = mmm_point_data_endpoint("temperature", "historical", lat, lon, "july")
 
+    all = mmm_point_data_endpoint("temperature", "historical", lat, lon)
+    temp_plate['historical']['all'] = all['historical']
+
+    jan = mmm_point_data_endpoint("temperature", "historical", lat, lon, "jan")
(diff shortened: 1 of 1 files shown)


cstephen (changes requested):
This PR is working as expected and the code looks good. However, I have a couple suggestions regarding the output:

- [x] I'd recommend using Flask's `jsonify` function instead of `json.dumps`. They do almost the same thing in this context, but the `jsonify` function also changes the HTTP header of the response to `application/json` instead of `text/html`. This does not appear to affect the web app, but it was preventing my browser from parsing and formatting the JSON output.

- [ ] Currently, the web app code uses the position within the JSON output array to determine what data belongs to which era (and month in the case of temperature). It'd be better if the API communicated this information in the JSON output itself. I was thinking the output could look closer to the following, maybe:

Precipitation:

```
{
  "historical": {
    "prmax": 23,
    "prmean": 14,
    "prmin": 9
  },
  "2010-2039": {
    "prmax": 26,
    "prmean": 16.4,
    "prmin": 10
  },
  "2040-2069": {
    "prmax": 28,
    "prmean": 17.6,
    "prmin": 11
  },
  "2070-2099": {
    "prmax": 32,
    "prmean": 19,
    "prmin": 10
  }
}
```

Temperature:

```
{
  "historical": {
    "all": {
      "tasmax": -1.8,
      "tasmean": -4.4,
      "tasmin": -7.2
    },
    "jan": {
      "tasmax": -3.1,
      "tasmean": -23.3,
      "tasmin": -40.1
    },
    "july": {
      "tasmax": 21.7,
      "tasmean": 13.5,
      "tasmin": 6.2
    }
  },
  "2010-2039": {
    "all": {
      "tasmax": 2,
      "tasmean": -2.3,
      "tasmin": -6.2
    },
    "jan": {
      "tasmax": -4.7,
      "tasmean": -20.6,
      "tasmin": -47.7
    }
    "july": {
      "tasmax": 28.7,
      "tasmean": 15.2,
      "tasmin": 1.9
    }
  },
  ...
}
```

This means we would need to go back and fix up the web app code accordingly, but I think this has the potential to simplify the web app code too.

PR #196, opened 2022-06-03
Handle missing Rasdaman coverages

charparr:
This PR provides a more general handling of missing Rasdaman coverages. The motivation for this PR was to prevent changes within and between the various Rasdaman servers (Zeus, Apollo) from blocking development on the API. In this PR an exception provides a specific warning message if the DescribeCoverage request to Rasdaman fails.

To test this behavior:

 - Point the API to Apollo
     - `export API_RAS_BASE_URL=http://apollo.snap.uaf.edu:8080/rasdaman/`
 - Edit Line 34 on `alfresco.py` such that coverage name for relative flammability is incorrect and doesn't match what is on Apollo, e.g., change `alfresco_relative_flammability_30yr` to `FOOalfresco_relative_flammability_30yr`
 - Run the API
 - Confirm that you observe a terminal error message, it should read: "Warning: Coverage 'FOOalfresco_relative_flammability_30yr' is missing from the Rasdaman server http://apollo.snap.uaf.edu:8080/rasdaman/ you are using."
 - Endpoints that hit this coverage will hit a TypeError (try http://localhost:5000/alfresco/flammability/point/65.0628/-146.1627) but other endpoints should work as expected. Failing with that error isn't ideal and it should probably 404 instead, but we'll need to deal with that on a per-endpoint basis and is probably beyond the scope of this PR.

### DIFF
--- a/fetch_data.py
+++ b/fetch_data.py
@@ -358,21 +358,27 @@ def parse_meta_xml_str(meta_xml_str):
 
 
 async def get_dim_encodings(cov_id):
-    """Get the dimension encodings from a rasdaman
-    coverage that has the encodings stored in an
-    "encodings" attribute
+    """Get the dimension encodings that map integer values to descriptive strings from a
+    Rasdaman coverage that stores the encodings in a metadata "encodings" attribute. We
+    handle exceptions where the coverage we are requesting encodings from does not exist
+    on the backend to prevent Rasdaman work from blocking API development.
 
     Args:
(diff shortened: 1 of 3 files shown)


cstephen (approved):
I tested this with different combinations of missing coverages, pointed at both Apollo and Zeus. It behaved as expected in all cases, and the coverages that existed continued to work. This is a great improvement!

PR #200, opened 2022-06-14
Modify snow endpoint to use mm units with no decimal places.

charparr:
This endpoint is being updated to return SI units with appropriate levels of precision. In this case, this means changing inches to millimeters with no decimal places.

To test this endpoint visit http://localhost:5000/mmm/snow/snowfallequivalent/hp/61.5/-147 and confirm that the results are as follows:

```
{
  "historical": {
    "sfemax": 4876, 
    "sfemean": 3701, 
    "sfemin": 2946
  }, 
  "projected": {
    "sfemax": 3945, 
    "sfemean": 3419, 
    "sfemin": 2589
  }
}
```

You may also try a few other locations - all returned values should have no decimal places.

XREF these issues:
https://github.com/ua-snap/rasdaman-ingest/issues/35
https://github.com/ua-snap/rasdaman-ingest/issues/34

Note - the "Prettier" formatter on VSCode did a lot of formatting to the HTML that documents this endpoint.

### DIFF
--- a/routes/snow.py
+++ b/routes/snow.py
@@ -24,7 +24,7 @@ from config import WEST_BBOX, EAST_BBOX
 
 snow_api = Blueprint("snow_api", __name__)
 # rasdaman targets
-sfe_coverage_id = "mean_annual_snowfall"
+sfe_coverage_id = "mean_annual_snowfall_mm"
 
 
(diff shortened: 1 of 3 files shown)


BobTorgerson:
This all looks good to me!

PR #203, opened 2022-06-14
Round degree days to whole numbers, and MAGT to one decimal

### LINKED ISSUE #202: Round all degree days and design index values to whole numbers, and MAGT values to 1 decimal place

cstephen:
Closes #202.

To test:

- Try out some example URLs from the [Design Index](http://localhost:5000/design_index/point) and [Degree Days](http://localhost:5000/mmm/degree_days/) endpoints (+ try some lat/lon coordinates of your own). Every value returned from these endpoints should be rounded to a whole number.
- Try the [Permafrost](http://localhost:5000/permafrost/point/) endpoint with several different lat/lon coordinates. Make sure all MAGT values are rounded to one decimal place.
- Observe that values in the example output in the documentation pages linked above match the expected decimal precision.

### DIFF
--- a/routes/degree_days.py
+++ b/routes/degree_days.py
@@ -257,7 +257,7 @@ def package_dd_point_data(point_data, var_ep, horp):
             year_index = 0
             for value in v_li:
                 if year in years:
-                    point_pkg[model][years[year_index]] = {"dd": value}
+                    point_pkg[model][years[year_index]] = {"dd": round(value)}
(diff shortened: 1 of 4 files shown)


charparr (approved):
I ran degree and permafrost queries for quite a few locations and they all returned properly rounded values (no decimals for degree days, one decimal for temperatures). The updates to the documentation are also correct.

PR #210, opened 2022-08-09
Sea ice concentration API endpoints

BobTorgerson:
This PR adds new endpoints to the SNAP API that are used to create a JSON output with key-value pairs for each year and month in the Rasdaman coverage 'hsia_arctic_production' along with min, mean, and max data for any point queried.

These endpoints are available at:

http://localhost:5000/mmm/seaice
http://localhost:5000/seaice/point

For testing this, be sure to set Apollo as your Rasdaman server of choice:

export API_RAS_BASE_URL=https://apollo.snap.uaf.edu/rasdaman


cstephen (changes requested):
Overall this works great and is a great addition to the API! But I found a few issues:

- [x] If you query an inland lat/lon point for any of the new sea ice endpoints (examples: `/mmm/seaice/64.78/-148.21`, `/mmm/seaice/64.78/-148.21/1900/2000`, `/seaice/point/64.78/-148.21`), I would expect the API to return a no data page, similar to ocean lat/lon queries from the temperature mmm endpoint for example: `/mmm/temperature/jan/historical/56.91/-146.51`. Instead the API appears to return lots of 120 values. I assume 120 is a nodata value that isn't being detected.
- [x] When `/hsia` is appended to the a  seaice query (example: `/seaice/point/62.91/-167.48/hsia`, an HTTP 500 server error is returned. Looks like the `/hsia` version of the endpoint is attempting to return a list. The Flask error in the terminal says:

  > The return type must be a string, dict, tuple, Response instance, or WSGI callable, but it was a list.
- [ ] Sea ice concentration is not mentioned from the "Temperature, Precipitation, Snow, and Degree Days Min, Mean, and Max" link text on the API front page.
- [x] Sea ice concentration is not mentioned in the intro text (under "Historical and Projected Minimum, Mean, and Maximum") or the text under "Service endpoints" on the `/mmm` page.
- [x] There's a minor indentation issue in the first block of example output on the `/mmm/seaice` page.

BobTorgerson:
@cstephen I believe all of these suggested changes have been resolved! Please continue your review at your convenience.

cstephen (approved):

brucecrevensten (changes requested):
OK, I pondered this and I think that my sense of unease is that we've got 2 endpoints for the full time series, one is used by the webapp and one isn't because the data model doesn't fit.  I think it makes sense to just offer 1 data endpoint which fits with the app we've built (and yeah, it'll require changes on the client side).  One insight is that the nested JSON approach isn't a great fit here -- essentially we've got an array of key/value pairs (YYYY-MM / %Conc), which is unlike the other nested JSON the API returns for other things (where each value has multidimensional identifiers).

I think what we want to offer here are 3 endpoint options:

 1. The `mmm` endpoint is fine as it is
 2. Modify the plain endpoint to return a data structure that our existing client can consume (it'll need more modification on the client side but I think that's worthwhile) -- probably an array like `[ { 'YYYY-MM': '%Conc' } ... ]`
 3. CSV output format, two columns (YYYY-MM / %Conc), proper header like is done elsewhere with place etc.

templates/seaice/seaice.html
```diff
@@ -0,0 +1,36 @@
...
+    "March": 98,
+    ...
+  }
+  ...
+}
+</pre>
+<b>The above output is structured like this:</b>
+<pre>
+{
+  &ltyear>: {
```
brucecrevensten, line 30: Probably remove this section, but I am morbidly curious why I'm seeing `&ltyear>` here (sans-semicolon...) looks right in browser, but I'd definitely recommend sticking with `&lt; ... &gt;` or (extreme/special case , probably never do this) `<![CDATA[ ... ]]>` ...

validate_data.py
```diff
@@ -86,6 +87,8 @@ def prune_nodata_list(data):
             pruned_value = prune_nodata(value)
             if len(pruned_value) > 0:
                 pruned.append(pruned_value)
+        else:
```
brucecrevensten, line 90: Little worried about this, would this be adding a kind of wildcard to what is accepted --?  (Kinda hard for me to read/understand in context tho!)

validate_data.py
```diff
@@ -117,8 +120,7 @@ def nullify_and_prune(data, endpoint):
 def postprocess(data, endpoint, titles=None):
     """Nullify and prune data, add titles, and return 404 if appropriate"""
     pruned_data = nullify_and_prune(data, endpoint)
-
-    if pruned_data in [{}, None, 0]:
+    if pruned_data in [{}, None, 0] or None in pruned_data:
```
brucecrevensten, line 123: Same thought as prior comment, I can't quite see what cases this covers!

brucecrevensten (changes requested):
A few things to change but overall very close, need to check in with folks on if `mmm` should go here.

routes/seaice.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    # initialize the output dict
+    di = dict()
+    for i in range(len(seaice_resp)):
+        di[f"{1850 + floor(i / 12)}-{str(i%12).zfill(2)}"] = seaice_resp[i]
+
+    return di
+
+
+def package_mmm_seaice_data(seaice_resp, start_year=None, end_year=None):
```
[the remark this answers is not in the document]
charparr: I've looked at this and I see what you mean, Bruce. The way I'm understanding this is that the `mmm` values are for all 12 months - so if you looked at the entire time series (all years) these summary values would be useful - but for a single year, yeah you are just getting the seasonal range, right? I'm not sure a mean concentration value for a year is useful. However, I also might not know what I'm talking about.

However, the monthly summary statistics would be very much in line with what I've seen in the limited amount of sea ice literature I've encountered.

routes/seaice.py
```diff
@@ -0,0 +1,214 @@
...
+        CSV response object
+    """
+
+    fieldnames = [
+        "year",
+        "month",
+        "concentration"
+    ]
+
+    months = [
```
charparr, line 48: No need to change it, but I just realized from working on something else that you can do:
```python
import calendar
months = list(calendar.month_name)
```
to generate a list of month names.

templates/seaice/seaice.html
```diff
@@ -0,0 +1,36 @@
...
+<p>
+  Example:
+  <a href="/seaice/point/63.73/-166.32"
+    >/seaice/point/63.73/-166.32</a
+  >
+</p>
+<b>Results from the queries above will look like this:</b>
+<pre>
+{
+  "1850": {
```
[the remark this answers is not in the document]
charparr: Consider adding a "Source Data" block like in https://earthmaps.io/mean_annual_precip/point/
Also, consider clarifying that returned values are percentages if that is the case.

charparr:
Adding Sea Ice to the API is a big step forward! This is awesome. I'm working in parallel on the landfast sea ice data right now and hope to land some PRs for that soon.

How do we want all the sea ice data to live together, if at all, in the API?

They can be split up entirely - so `/seaice/point` would be independent from `/landfastice/point` in which case there is nothing further to do here.

Or, does something like `/seaice/concentration/point` and `/seaice/landfast/point` seem better? Does it matter? I don't have strong preferences right now but have been thinking in ice mode lately

Pipfile
```diff
@@ -20,6 +20,7 @@ shapely = "*"
 geos = "*"
 rtree = "*"
 pygeos = "*"
+black = "*"
```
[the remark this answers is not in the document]
BobTorgerson: This has been moved into development packages rather than production packages.

routes/seaice.py
```diff
@@ -0,0 +1,214 @@
...
+    Arguments:
+        seaice_resp -- the response(s) from the WCS GetCoverage request(s).
+
+    Returns:
+        di -- a nested dictionary of all SFE values
+    """
+    # initialize the output dict
+    di = dict()
+    for i in range(len(seaice_resp)):
+        di[f"{1850 + floor(i / 12)}-{str(i%12).zfill(2)}"] = seaice_resp[i]
```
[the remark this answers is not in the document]
BobTorgerson: This has been changed to have January = 01 and December = 12 for the returned JSON data.

templates/seaice/abstract.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'base.html' %}
+{% block content %}
+<h3>Pan Arctic Sea Ice Concentration</h3>
```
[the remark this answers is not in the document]
BobTorgerson: Done.

templates/seaice/seaice.html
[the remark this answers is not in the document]
BobTorgerson: Added the Source Data block to point at the NSIDC-0051 dataset

templates/seaice/seaice.html
BobTorgerson: Cleaned up these and ensured they had semi-colons with them

validate_data.py
BobTorgerson: Your Spidey-sense was good on this Bruce! This had been for a special case in which we returned the flat list from Rasdaman which has since been removed! Thanks for catching this!

validate_data.py
BobTorgerson: Same thing as the previous one! This was for a now defunct use case.

routes/seaice.py
brucecrevensten: Glad to know this exists -- I always consider it kind of a "bad smell" when I enumerate months in code!  I'd suggest changing it just to remove the hard coded months.

routes/seaice.py
```diff
@@ -0,0 +1,215 @@
...
+def about_seaice():
+    return render_template("seaice/abstract.html")
+
+
+@routes.route("/seaice/point/")
+def about_seaice_point():
+    return render_template("seaice/point.html")
+
+
+@routes.route("/mmm/seaice/")
```
brucecrevensten, line 138: I think I'd still recommend removing the mean/min/max endpoint for this variable.

templates/seaice/point.html
```diff
@@ -0,0 +1,51 @@
+{% extends 'base.html' %} {% block content %}
+<h3>Pan&ndash;Arctic Sea Ice Concentration Point Query</h3>
+<p>
+  This endpoint accesses historical monthly data on sea ice concentration in the
+    Arctic from 1850-2021.
```
brucecrevensten, line 5: So so minor, but ideally this would be `&ndash;` ... only worth changing to get in the habit of looking at dashes and stuff with a careful eye when writing anything in HTML!

templates/seaice/point.html
```diff
@@ -0,0 +1,51 @@
...
+    <th>Gridded Historical Dataset</th>
+    <th>Historical Date Range</th>
+    <th>Spatial Resolution</th>
+    <th>Spatial Coverage</th>
+  </tr>
+</thead>
+<tbody>
+  <tr>
+    <td><a href="https://nsidc.org/data/nsidc-0051/versions/1">Sea Ice Concentrations from Nimbus-7 SMMR and DMSP SSM/I-SSMIS Passive Microwave Data, Version 1</a></td>
+    <td>1850 - 2022</td>
```
brucecrevensten, line 45: This could be (wait for it... :) ) `1850&ndash;2022`

brucecrevensten (changes requested):
Looks good with only some very minor comments; I do still recommend removing the min/mean/max endpoint though -- it's not used by any app, and unless it's modified to only examine stacks of specific months, I'm not sure how useful it is for this specific variable.  I think it'd make sense to add this back if we have a client app that uses it, though, and maybe that's something we'd want to talk to John about with a future revision of the Atlas app?

brucecrevensten (approved):

PR #216, opened 2022-09-12
Landfast Sea Ice Endpoint 1.0

charparr:
This PR adds the first version of the landfast sea ice endpoint.

The data accessed by this endpoint are described in this [curation notebook](https://github.com/ua-snap/ardac-curation/blob/main/seaice/pipeline/qc.ipynb) and have been through a [QC process](https://github.com/ua-snap/ardac-curation/blob/main/seaice/pipeline/qc.ipynb). The data rely on the coverage `landfast_sea_ice_extent` which is currently on Apollo, but not Zeus.

We expect this service endpoint to significantly grow in both query versatility and in availability of diffferent data variables - but this initial service only provides a time series of landfast ice status (present or absent) for a geographic point. A value of `1` indicates that landfast ice exists. A value of `0` indicates that landfast ice does not exist.

To test this PR:
 - Point your local API at Apollo
 - Visit http://localhost:5000/landfastice/ and /landfastice/point/ and verify that text is clear and complete and that example links work as you'd expect.
 - Verify that "Landfast Sea Ice" is present on the API home page.
 - Try a few different points for queries and CSV results. You may need to scroll through the time series or sort it to see a scatter of `0` and `1` values.
 - Try a terrestrial lat-lon (like 65/-147) and verify you get a "No Data" message.

Other notes:
 - This PR adds a function where you can scrape between known tags in the XML returned by the DescribeCoverage request to Rasdaman. This is done to fetch and create the time index.

### DIFF
--- a/fetch_data.py
+++ b/fetch_data.py
@@ -323,11 +323,13 @@ def get_from_dict(data_dict, map_list):
 
 def parse_meta_xml_str(meta_xml_str):
     """Parse the DescribeCoverage request to get the XML and
-    restructure to dict
+    restructure the block called "Encoding" to a dict.
 
-    Args:
-        meta_xml_str (str): decoded text XML response from
-            DescribeCoverage WCS query
+    Arguments:
+        meta_xml_str (str): string representation of the byte XML response from the WCS DescribeCoverage request
+
+    Returns:
+        dim_encodings (dict): lookup table to match data axes or parameters to integer encodings, e.g., '2': 'GFDL-CM3'
     """
     meta_xml = ET.ElementTree(ET.fromstring(meta_xml_str))
     # wow xml
@@ -357,23 +359,45 @@ def parse_meta_xml_str(meta_xml_str):
     return dim_encodings
 
 
-async def get_dim_encodings(cov_id):
+def get_xml_str_between_tags(meta_xml_str, tag):
+    """Get string encapsulated by a known XML tag.
+
+    Arguments:
+        meta_xml_str (str): string representation of the byte XML response from the WCS DescribeCoverage request
+        tag (str): the xml string that encapsulates the desired information, e.g., 'gmlrgrid:coefficients'
+
+    Returns:
+        str_within_tag (str): string encapsulated by the provided XML tag
+    """
+    tag_open = f"<{tag}>"
+    tag_close = "<"
+    str_after_tag = meta_xml_str.split(tag_open)[1]
+    str_within_tag = str_after_tag.split(tag_close)[0]
+    return str_within_tag
+
+
+async def get_dim_encodings(cov_id, scrape=None):
     """Get the dimension encodings that map integer values to descriptive strings from a
     Rasdaman coverage that stores the encodings in a metadata "encodings" attribute. We
     handle exceptions where the coverage we are requesting encodings from does not exist
-    on the backend to prevent Rasdaman work from blocking API development.
+    on the backend to prevent Rasdaman work from blocking API development. We can use the same request to scrape various other parts of the DescribeCoverage XML response, but this optional.
 
     Args:
         cov_id (str): ID of the rasdaman coverage
+        scrape (2-tuple of strings): (description, tag to scrape between)
 
     Returns:
-        dict of encodings, with axis name as keys holding
-        dicts of integer-keyed categories
+        dim_encodings (nested dict): a lookup where coverage axis names are keys that store dicts of integer-keyed categories.
     """
     meta_url = generate_wcs_query_url(f"DescribeCoverage&COVERAGEID={cov_id}")
     try:
         meta_xml_str = await fetch_data([meta_url])
         dim_encodings = parse_meta_xml_str(meta_xml_str)
+        if scrape is not None:
+            scrape_desc, scrape_tag = scrape
+            dim_encodings[scrape_desc] = get_xml_str_between_tags(
+                meta_xml_str, scrape_tag
+            )
         return dim_encodings
     except:
         print(
--- a/fetch_data.py
+++ b/fetch_data.py
@@ -378,9 +378,7 @@ def get_xml_str_between_tags(meta_xml_str, tag):
 
 async def get_dim_encodings(cov_id, scrape=None):
     """Get the dimension encodings that map integer values to descriptive strings from a
-    Rasdaman coverage that stores the encodings in a metadata "encodings" attribute. We
-    handle exceptions where the coverage we are requesting encodings from does not exist
-    on the backend to prevent Rasdaman work from blocking API development. We can use the same request to scrape various other parts of the DescribeCoverage XML response, but this optional.
+    Rasdaman coverage that stores the encodings in a metadata "encodings" attribute. We handle exceptions where the coverage we are requesting encodings from does not exist on the backend to prevent Rasdaman work from blocking API development. We can use the same request to scrape various other parts of the DescribeCoverage XML response, but this optional.
 
     Args:
         cov_id (str): ID of the rasdaman coverage
--- a/routes/landfastice.py
+++ b/routes/landfastice.py
@@ -18,6 +18,7 @@ from validate_request import (
     validate_latlon,
     project_latlon,
 )
+from validate_data import postprocess
 from . import routes
 from config import WEST_BBOX, EAST_BBOX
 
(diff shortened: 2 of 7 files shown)


fetch_data.py
```diff
@@ -357,23 +359,45 @@ def parse_meta_xml_str(meta_xml_str):
...
+    str_within_tag = str_after_tag.split(tag_close)[0]
+    return str_within_tag
+
+
+async def get_dim_encodings(cov_id, scrape=None):
     """Get the dimension encodings that map integer values to descriptive strings from a
     Rasdaman coverage that stores the encodings in a metadata "encodings" attribute. We
     handle exceptions where the coverage we are requesting encodings from does not exist
-    on the backend to prevent Rasdaman work from blocking API development.
+    on the backend to prevent Rasdaman work from blocking API development. We can use the same request to scrape various other parts of the DescribeCoverage XML response, but this optional.
```
cstephen, line 383: Minor nitpick, but this and other comments could use some manual word-wrapping.

cstephen (changes requested):
Things are looking good! I've made a few small suggestions in the code comments. I also  tried a handful of inland queries and was surprised that the following two coordinates from northern Alaska return data:

http://127.0.0.1:5000/landfastice/point/68.64/-158.02/
http://127.0.0.1:5000/landfastice/point/68.27/-144.36/

Any idea what's going on there?

templates/landfastice/point.html
```diff
@@ -0,0 +1,64 @@
...
+  <a href="/landfastice/point/70.6123/-149.8869?format=csv"
+    >/landfastice/point/70.6123/-149.8869?format=csv</a
+  >
+</p>
+<h4>Source data</h4>
+<!-- This is a bit of a placeholder until the we launch ARDAC or get PI approval to publish in the Catalog.-->
+<p>
+  These data are part of the Arctic Data Collaborative (ARDAC). Scientists
+  working on Arctic research need easily obtainable climate and environmental
+  data. ARDAC is a system of online technologies that curate distribute relevant
```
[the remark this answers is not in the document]
charparr: Fixed!

templates/landfastice/abstract.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'base.html' %} {% block content %}
+<h3>Landfast Sea Ice Extent.</h3>
```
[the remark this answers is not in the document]
charparr: Good catch. Fixed.

routes/landfastice.py
```diff
@@ -0,0 +1,139 @@
...
+    """Create CSV file with metadata string and location based filename.
+    Args:
+        data_pkg (dict): JSON-like object of data
+        lat: latitude for points or None for polygons
+        lon: longitude for points or None for polygons
+    Returns:
+        CSV response object
+    """
+    fieldnames = [
+        "time",
```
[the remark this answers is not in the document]
charparr: Agreed - it definitely is not time. Fixed.

fetch_data.py
charparr: Thanks!! Yeah I think had a return in there that broke the wrapping. This should be fixed now but please double check (it looked ok in the diff to me and in VS Code).

charparr:
http://127.0.0.1:5000/landfastice/point/68.64/-158.02/
http://127.0.0.1:5000/landfastice/point/68.27/-144.36/

Should now 404 as expected. I've gone ahead and nullified the nodata (0 values).
In my head that was going to break the continuity of the time series, but it doesn't! So there is no reason to not do this.

Some backend changes to this dataset are likely at some point in the future, so we may add some other post processing or no data values as needed.

cstephen:
Thanks for fixing those URLs! I'm still seeing some occasional `1` values at locations that are pretty far inland, however. For example:

http://127.0.0.1:5000/landfastice/point/70.48/-156.44/
http://127.0.0.1:5000/landfastice/point/69.34/-160.35/

These return mostly `null` values but there are some `1`s in there too. I'm not as familiar with these data as you, but I'm thinking we should maybe mask these inland values out during post-processing? If that's already part of the plan for future work, feel free to ignore!

Also, the mixture of `1`s and `null`s at valid locations feels a bit off to me. `1`s and `0`s seem more appropriate since this is binary information. I'm hoping we can use a different nodata value like 255 or -9999 for invalid/masked locations and keep `0`s in the output for valid/unmasked locations.

fetch_data.py
cstephen: No luck on the word-wrapping, unfortunately. Or we might be using the term word-wrapping in different ways. IMO, we should add more line breaks to try to keep comment lines shorter than ~100 characters.

cstephen: We decided that it's perfectly acceptable to have long docstring with no manual line breaks, and it passes Black formatting. So feel free to disregard this comment!

charparr:
Hmm so for the first location the offending data values occur for the dates/GeoTIFFS in the range `6-18-2008` through `7-12-2008`, and for the second location they occur for `06-29-2008` through `07-14-2008`. I'm wondering if these data were never actually used by the scientists and left in a not-fully-processed state. Consider this quote from the most recent [paper](https://doi.org/10.1016/j.coldregions.2014.03.003) that Hajo sent (unseen by me at time of the initial pre-processing work):

>...Also, Radarsat data became a commercial product after April 2008 with data costs that prohibited acquiring complete coverage of our region. Instead, we acquired Envisat Advanced Synthetic Aperture Radar (ASAR) data to capture break-up for 2008.

My instinct here is just truncate the time series for this Radarsat dataset and see if that fixes the terrestrial query issue - if not, there is likely a pre-processing issue and/or the initial effort to characterize the variability of the dataset missed something.

charparr:
The landfast sea ice data has a relatively fine spatial resolution (100 m) and hugs the coast tightly, and in some places occupies an intertidal zone. Performing a terrestrial mask on the data is virtually impossible because any coastline geometry is static while the coast is dynamic, and so even if the coastal geometry can capture the full complex shape of the landfast sea ice data, there will be conflicts where the landfast sea ice is present in an area partially or completely identified as land by the coastal geometry. We could of course edit or create a custom coastal mask using the maximum extent of the landfast sea ice, but that feels circular: using the ice data to delineate the coast, which is used to then mask the ice data.

We could mask out some "core" area of Alaska's land surface, but some land area would always exist outside that area near the coast, and so you'd have some terrestrial areas that are "no data" , and some terrestrial areas that are a sequence of "no landfast ice" values.

What I've done is truncate the time series at `04/30/2008` to reflect what we learned from Hajo, namely that after April 2008 a different data source / product was used.

http://127.0.0.1:5000/landfastice/point/70.38/-156.44/
and
http://127.0.0.1:5000/landfastice/point/69.34/-160.35/
now return a "No data" template.

routes/landfastice.py
```diff
@@ -0,0 +1,139 @@
...
+        "date",
+        "value",
+    ]
+    csv_dicts = build_csv_dicts(
+        data_pkg,
+        fieldnames,
+    )
+    metadata = "# Landfast Ice Status: 0 indicates absence and 1 indicates presence.\n"
+    filename = "Landfast Ice Extent for " + lat + ", " + lon + ".csv"
+    return write_csv(csv_dicts, fieldnames, filename, metadata)
```
BobTorgerson, line 89: The CSV that is returned does not have 0's for the absence of data at a given time, but rather is a blank space. This is due to the null that is going into the CSV due to the pruning that happens before creating the CSV. We could either change the CSV metadata to start that a blank space represents absence in the data or we will want to fix this to have 0 show up in the CSVs.

BobTorgerson (changes requested):
Things look and work well! I noted the CSV difference between what was written in the header of the CSV and the actual structure of the CSV, and that there appears to be a conditional and function parameter that can be removed as the code stands now.

routes/landfastice.py
```diff
@@ -0,0 +1,139 @@
...
+        ice_season (str): Ice season string (YYYY) between 1996 and 2007 or 'all' for the entire time series.
+
+    Returns:
+        dt_range (pandas DatetimeIndex): a time index with daily frequency
+    """
+
+    if ice_season == "all":
+        timestamps = [x[1:-2] for x in landfastice_encodings["time"].split(" ")]
+        date_index = pd.DatetimeIndex(timestamps)
+    else:
```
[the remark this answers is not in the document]
charparr: Ah thanks Bob! That's leftover from some time slicing experimentation that I neglected to remove. I'll pull this for now. There will be some kind of slicing and dicing in the future, but we don't know how or why yet.

routes/landfastice.py
charparr: I'll make the metadata change unless there are strong preferences for re-populating with a `0` value.

cstephen (approved):
I've verified that the four inland URLs I mentioned before are returning 404 pages as expected:

http://127.0.0.1:5000/landfastice/point/68.64/-158.02/
http://127.0.0.1:5000/landfastice/point/68.27/-144.36/
http://127.0.0.1:5000/landfastice/point/70.48/-156.44/
http://127.0.0.1:5000/landfastice/point/69.34/-160.35/

Everything seems to be working as expected (aside from the CSV issue @BobTorgerson mentioned). I personally have a strong preference for returning the data as `1`s and `0`s (or maybe even `true` and `false`) instead of `1`s and `null`s, but this is an aesthetic preference and not a showstopper, and also something we can iterate on later if needed. Marking this as approved!

BobTorgerson (approved):
Thanks for making those changes Charlie! This all looks good to merge to me.

PR #229, opened 2022-11-04
New endpoint for generating combined JSON for all Arctic EDS report sections

BobTorgerson:
This PR adds the new endpoint /eds/all/<lat>/<lon> to the API. This combines all of the report sections in the Arctic EDS web application and generates a long JSON listing of the returned data.

Each report section has a key indicating its underlying data such as "thawing_index" or "precipitation". 

Try a few valid lat / lon coordinates such as: 
http://localhost:5000/eds/all/68.0764/-153.5501
http://localhost:5000/eds/all/62.0764/-143.5501

Try some invalid lat / lon coordinates to see what is returned as well, such as:
http://localhost:5000/eds/all/22.0764/-123.5501
http://localhost:5000/eds/all/52.0764/-123.5501

### DIFF
--- a/routes/eds.py
+++ b/routes/eds.py
@@ -15,6 +15,19 @@ eds_api = Blueprint("eds_api", __name__)
 
 
 async def run_fetch_all_eds(lat, lon):
+    """
+    Fetches a list of data API end points to generate a multiple
+    variable JSON that can be used to generate a full report in the
+    Arctic EDS website.
+
+    Args:
+        lat (float): latitude
+        lon (float): longitude
+
+
+    Returns:
+        Multiple variable JSON object
+    """
     host = request.host_url
     all_requests = [
         f"{host}eds/temperature/{lat}/{lon}",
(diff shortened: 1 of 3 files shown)


brucecrevensten (approved):

routes/eds.py
```diff
@@ -0,0 +1,115 @@
...
+        1 - Precipitation
+        2 - Snowfall
+        3 - Design Freezing Index
+        4 - Design Thawing Index
+        5 - Freezing Index
+        6 - Heating Degree Days
+        7 - Thawing Index
+        8 - Geology
+        9 - Physiography
+       10 - Permafrost
```
cstephen, line 112: I don't think this comment applies anymore, so I suggest just removing it since the results are self-explanatory now.

cstephen (changes requested):
Functionality looks good, and all of my test queries finished in under 40 seconds. Nice work!

My only suggestion is to remove the "Order of returned JSON" comment since it no longer applies. I'll get this merged once that is fixed.

routes/eds.py
BobTorgerson: Good point! I've removed that comment and it's good to merge.

cstephen (approved):

PR #232, opened 2022-11-22
Updates caching for the API with the beetles API section + the Arctic EDS endpoint for all communities.

BobTorgerson:
This PR adds the calls to fill the cache with the beetle endpoints for area and point queries, along with the Arctic EDS community endpoints to cache all identified communities.

To test this, you will want to comment out lines 82, 85 and 86 of routes/recache.py: https://github.com/ua-snap/data-api/blob/40ae48eab898b0f9d1814f51b564eca9147514c8/routes/recache.py#L82-L86

and add this to line 80 of that file:

print(url)

This is a roundabout way of identifying what would be called on the remote server when recaching. Ensure you see the beetles endpoint and the ArcticEDS endpoint shown in the long listing of all API calls in the console.


cstephen (approved):
I followed the testing instructions in the PR description and confirmed that the following types of URLs show up in the list of URLs to cache:

- beetle points
- beetle areas
- "all" Arctic EDS queries

Everything is working as expected.

PR #239, opened 2022-12-15
Add service endpoints for the most recent GIPL 2.0 outputs

charparr:
This PR adds service endpoints for the most recent of GIPL model outputs (ten variables at 1 km spatial resolution) that were provided to SNAP by Sergey Marchenko in August 2022. The backend data supporting these endpoints is a Rasdaman coverage called `crrel_gipl_outputs` (currently on Apollo only) that was created by ingesting 6000 GeoTIFFs. The spatial extent of these data are Alaska only - although the Western Aleutian chain is not present. 

There are 3 model cases (5-Model-Average, GFDL-CM3, and NCAR-CCSM4), and two scenarios (RCPs 4.5 and 8.5).

Only point queries are implemented at this time.

To test this PR, try queries constructed as follows to fetch the complete set of GIPL model outputs (all models, scenarios, and years) for a point location:

http://localhost:5000/permafrost/point/gipl/66.0628/-146.1627

To slice a time span from these annual-frequency data, simply append a start and stop year to your path - try these examples to fetch data and test validation:

http://localhost:5000/permafrost/point/gipl/65/-148/2040/2049
http://localhost:5000/permafrost/point/gipl/65/-148/2049/2040
http://localhost:5000/permafrost/point/gipl/65/-148/2040/3049
http://localhost:5000/permafrost/point/gipl/65/-148/1040/2049

To slice a time span from these annual-frequency data and then compute mean-min-max summaries for each of the ten variables, simply append "mmm" to your URL:

http://localhost:5000/permafrost/point/gipl/65/-148/2040/2069/mmm

To test validation:

http://localhost:5000/permafrost/point/gipl/65/-148/2040/2069/foo

Highlights:
 - Check the new WCPS syntax courtesy of Dimitar! Pretty clean, and easy to flip between min / mean / max.
 - I like how this route came together: all data > time slice > time slice summarized

Notes:
 - I tread carefully to make sure the other permafrost data (including the Melvin et al 4 km GIPL outputs served used in NCR) endpoints were not impacted. But keep an eye out for that.
 - Also, take a close look at the template. Are the GIPL outputs distinct enough? I think we'd love to mothball the 4 km stuff, but that could be work because of the NCR.
 - In the code I've tried to distinguish the new GIPL data by using `gipl1km` everywhere.

Misses:
 - We need to add to the Data Source table. You have to scroll far to even get to that table, but it is important. This means adding the 6000 GeoTIFFs to our GeoNetwork Catalog.

### DIFF
--- a/fetch_data.py
+++ b/fetch_data.py
@@ -364,19 +364,20 @@ def parse_meta_xml_str(meta_xml_str):
     return dim_encodings
 
 
-def get_xml_str_between_tags(meta_xml_str, tag):
-    """Get string encapsulated by a known XML tag.
+def get_xml_str_between_tags(meta_xml_str, tag, occurrence=1):
+    """Get string encapsulated by a known XML tag. Use this function to retrieve time axis values that are not encapsulated by a dictionary and/or are not within the metadata 'Encoding' block.
 
     Arguments:
         meta_xml_str (str): string representation of the byte XML response from the WCS DescribeCoverage request
         tag (str): the xml string that encapsulates the desired information, e.g., 'gmlrgrid:coefficients'
+        occurrence (int): the occurrence of the tag to parse. some tags are repeated several times in the XML response
 
     Returns:
         str_within_tag (str): string encapsulated by the provided XML tag
     """
     tag_open = f"<{tag}>"
-    tag_close = "<"
-    str_after_tag = meta_xml_str.split(tag_open)[1]
+    tag_close = f"</{tag}>"
+    str_after_tag = meta_xml_str.split(tag_open)[occurrence]
     str_within_tag = str_after_tag.split(tag_close)[0]
     return str_within_tag
 
@@ -387,7 +388,7 @@ async def get_dim_encodings(cov_id, scrape=None):
 
     Args:
         cov_id (str): ID of the rasdaman coverage
-        scrape (2-tuple of strings): (description, tag to scrape between)
+        scrape (3-tuple): (description (str), tag to scrape between (str), and the occurrence (int) of the tag to search for)
 
     Returns:
         dim_encodings (nested dict): a lookup where coverage axis names are keys that store dicts of integer-keyed categories.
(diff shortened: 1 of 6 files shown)


cstephen (approved):
I kicked the tires on this using the URL examples provided in the PR description, plus:

- Each of the example URLs with different lat/lon coordinates, both inland and water.
- More bogus URL parameters (letters instead of numbers, invalid date ranges, etc.).
- Tested the min/mean/max values of year ranges. Longer year ranges that extended farther into the future produced higher max numbers, and different means, but not lower min numbers.

In all cases, the API behaved as expected. This all looks great to me!

Regarding the documentation template, I'm not sure what to do here either since we will presumably phase out the 4km permafrost data in the more-distant future. Maybe we could divide this page up with big headings that say "Permafrost point query (1km)" and "Permafrost point query (4km)". I defer to you on this though, @charparr.

The new WCPS query is impressive and will surely be a reference point for future complex WCPS query construction!

fetch_data.py
```diff
@@ -364,19 +364,20 @@ def parse_meta_xml_str(meta_xml_str):
     return dim_encodings
 
 
-def get_xml_str_between_tags(meta_xml_str, tag):
-    """Get string encapsulated by a known XML tag.
+def get_xml_str_between_tags(meta_xml_str, tag, occurrence=1):
```
brucecrevensten, line 367: Not saying do this now, but we probably want to just use XPath to extract these expressions now that there's a bit more complexity.  Either adapt this or make a ticket to do so.

brucecrevensten (commented):
General observation: were some files not-linted in a long time, or is there some kind of a lint-war going on here between different configurations?  (I didn't think Black really did that).

fetch_data.py
charparr: Hey Bruce I'll ticket this. The way this is being done now is certainly a bit hacky - but I think I'll need a bit of research to find the `lxml` or equivalent fix. Related: I'd propose that we move this function and ones like it to a `describe_data` module. We have a few pieces now that just describe the data and unwrap the coverage encodings - but don't actually fetch any data. I think having them in a common module would help us de-duplicate code because we'll be able to see spots to make those calls for flexible + general.

charparr: As to the linting - I have my editor set to format Python files with Black upon save. So I'm not sure if they hadn't been formatted in a while or what...that seems weird.

PR #255, opened 2023-04-06
Integrates HUC12 boundary calls and ALFRESCO local requests to Geoserver

BobTorgerson:
This PR changes the last of the remnants of our old process for downloading shapefiles locally and opening them via GeoPandas for analysis. This points the calls for the old HUC12 GDF to Geoserver WFS responses that are then loaded into a temporary GDF for analysis. In some cases, we are letting Geoserver do all of the work that was originally required such as finding the nearest HUC12 that intersects with a given lat / lon for the ALFRESCO local requests. 

Let me know if you have any questions, and please let me know if you spy anything that isn't right and should be changed.

Things to check:

- Ensure the ALFRESCO code works for any of the local endpoints for flammability and veg_change
- Ensure you don't see any outstanding areas where the GDFs are called from luts.py
- Any code that could be improved using your knowledge of Geopandas

### DIFF
--- a/routes/vectordata.py
+++ b/routes/vectordata.py
@@ -12,7 +12,6 @@ from shapely.geometry import Point, box
 # local imports
 from . import routes
 from luts import (
-    shp_di,
     all_jsons,
     areas_near,
 )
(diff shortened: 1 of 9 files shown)


routes/vectordata.py
charparr, line 1: All red :100

I don't think this import is being used any more either, thanks GeoServer :)
https://github.com/ua-snap/data-api/blob/feb4f21c4a60c92dd29e95748916a6f003c35f83/routes/vectordata.py#L10

And I think we can rip out these now-unused imports too:
https://github.com/ua-snap/data-api/blob/feb4f21c4a60c92dd29e95748916a6f003c35f83/routes/vectordata.py#L5-L7

And we no longer need `GS_BASE_URL` so it no longer needs to be fetched from the config
https://github.com/ua-snap/data-api/blob/feb4f21c4a60c92dd29e95748916a6f003c35f83/routes/vectordata.py#L18

validate_data.py
charparr, line 1: I don't think this needs to be changed now, but looking at this module I feel like the `place_name_and_type` and `get_poly_3338_bbox` are a weird fit with the other functions in this module, that well, validate data by pruning nodata values, checking for empty data packages, etc.

validate_request.py
```diff
@@ -12,7 +12,6 @@
 from config import GS_BASE_URL, WEST_BBOX, EAST_BBOX, SEAICE_BBOX
```
charparr, line 12:
```suggestion
from config import WEST_BBOX, EAST_BBOX, SEAICE_BBOX
```
I think GS_BASE_URL is also no longer used in this module.

charparr (changes requested):
This is a great PR. It reduces and simplifies the codebase quite a bit and helps light the path in terms of where to go next with improving the API and making our own lives easier. I've left a few comments about cleaning up some unused imports, but otherwise I think this is good to merge!

To test this PR I ran the API locally and hit several different polygon summary endpoints, and tested the ALFRESCO endpoints in particular. I also tried a few different queries to `places/search` like so: http://localhost:5000/places/search/65/-147 so that I could check out what is being found and in particular see the various HUCs.  Everything looks good to my eye!

kyleredilla (changes requested):
Agree with Charlie - these changes are exciting to see and really make the code easier to interpret.

I tested the ALFRESCO endpoints, including a bunch of queries to the HUC-12- and other polygon-based aggregation endpoints, and received no unexpected results.

In addition to Charlie's suggestions, I think my only other suggestion is that I think we can remove the section in the README about updating the data as well now, right? Maybe either include that in this PR, or make an issue for it?

kyleredilla (approved):

PR #256, opened 2023-04-26
Beetle coverage changes to match climate protection meaning of data

BobTorgerson:
This PR makes the changes required to match the beetle coverage data's meaning of climate protection from spruce beetles. This includes returning new fields in CSVs and the definitions being high climate protection, minimal climate protection, and no climate protection.

This PR is required for the changes to the NCR application for the beetle coverage portion of the report.

### DIFF
--- a/routes/beetles.py
+++ b/routes/beetles.py
@@ -47,10 +47,10 @@ dim_encodings = {
         2: "2070-2099",
     },
     "snowpack": {0: "low", 1: "medium"},
-    # The 0 in beetle_risk represents NO DATA in this context,
+    # The 0 in climate_protection represents NO DATA in this context,
     # but needs to remain as 0 to allow for the pruning function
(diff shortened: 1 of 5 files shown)


cstephen (changes requested):
Looks good and works as expected! The only issue I found is that a couple places weren't updated with the new risk vs. climate protection terminology:

- [ ] The metadata line for downloadable CSVs still says "Values shown are for risk level for spruce beetle spread in the area."
- [ ] On the documentation page, the explanation for `snowpack` says: "Relative snowpack used to model spruce beetle risk"

This branch looks mergeable once those are updated.

BobTorgerson:
@cstephen Thanks for catching those! It's hard to ensure you've caught all of the places where this was a thing due to how the wording changed for describing just about everything :) Please review again when you have the opportunity.

cstephen (approved):

Issue #261, opened 2023-05-12
Split summary of GIPL permafrost min, mean, and max into sections for each model


BobTorgerson:
All of the models are currently averaged together to generate the min, mean, and max for a given time period, but we want those separated into each model for a given time period.

brucecrevensten:
We may want to move forward with this, or, we may want to back off on implementing mean/min/max for permafrost.  Check with team before finalizing

BobTorgerson:
@charparr Do you have any idea if this is something that we want to move forward with? Is Nancy the right person to ask if this is important to do?

charparr:
@BobTorgerson good question! I think mean-min-maxing chunks of time by model is worth doing.

Here is why:
 - It is thematically consistent with other datasets
 - I'm sure somewhere (EDS?) users will want to see the model variance within a time period (i.e., which model is most variable for GIPL?)
 - We need to summarize time chunks, so we need to min-mean-max the GIPL data in some way no matter what. Individual years will run the gamut of cool/wet/hot/dry which can hide the trends. We saw this in a few of our early WMS sample requests.

brucecrevensten:
This ultimately is probably a question of, "what do we want the updated gipl2 endpoint to look like."  It's valid and we have all the data needed in the coverage, but can summarize/combine it differently in the endpoint.  We want to make the returned data more granular, and this is part of that.

brucecrevensten:
This work was completed as part of integrating the GIPL2 stuff, closing!

Issue #258, opened 2023-05-03
Update dataset links/docs for permafrost endpoint


brucecrevensten:
It's still referencing Melvin in the abstract endpoint, should be referencing CRREL

brucecrevensten:
Closed in other tickets in this milestone.

PR #262, opened 2023-05-12
GIPL Permafrost updates

BobTorgerson:
Updates GIPL Permafrost formatting for the summarized variables over a time range and creates an endpoint for gathering all of the time ranges desired by NCR. 

This PR will split out the summarized GIPL permafrost data out into each models min, mean, and maximum in the returned JSON.

You can see this in: http://localhost:5000/permafrost/point/gipl/65/-156/2021/2039/mmm

Additionally, this adds a new /ncr endpoint that combines all of the required data for the NCR Permafrost section from a single call to the API.

This can be seen here: http://localhost:5000/ncr/permafrost/point/65/-156

### DIFF
--- a/routes/permafrost.py
+++ b/routes/permafrost.py
@@ -148,12 +148,13 @@ def package_gipl1km_wcps_data(gipl1km_wcps_resp, package_type):
     summary_methods = ["min", "mean", "max"]
     if package_type == "historical":
         gipl1km_wcps_point_pkg["Historical"] = dict()
+        gipl1km_wcps_point_pkg["Historical"]["Historical"] = dict()
         for resp, stat_type in zip(gipl1km_wcps_resp[0], summary_methods):
-            gipl1km_wcps_point_pkg["Historical"][f"gipl1km{stat_type}"] = dict()
+            gipl1km_wcps_point_pkg["Historical"]["Historical"][f"gipl1km{stat_type}"] = dict()
             for k, v in zip(
                 gipl1km_historical_dim_encodings["variable"].values(), resp
             ):
-                gipl1km_wcps_point_pkg["Historical"][f"gipl1km{stat_type}"][k] = round(
+                gipl1km_wcps_point_pkg["Historical"]["Historical"][f"gipl1km{stat_type}"][k] = round(
                     v, 1
                 )
     else:
@@ -213,8 +214,9 @@ def package_gipl1km_point_data(gipl1km_point_resp, time_slice=None):
         for t in tx:
             year = t.date().strftime("%Y")
             gipl1km_point_pkg["Historical"][year] = dict()
+            gipl1km_point_pkg["Historical"][year]["Historical"] = dict()
             for gipl_var_name in gipl1km_historical_dim_encodings["variable"].values():
-                gipl1km_point_pkg["Historical"][year][gipl_var_name] = round(
+                gipl1km_point_pkg["Historical"][year]["Historical"][gipl_var_name] = round(
                     flat_list[i], 1
                 )
                 i += 1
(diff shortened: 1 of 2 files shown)


BobTorgerson:
Closes #261

charparr:
Now should close  Close #258

charparr:
We may need to take a look at the CSV output, the columns will probably have to be:
model - summary - variable - value

image

charparr (commented):
Bob this looks good! Thanks for cleaning this up. I took this opportunity to spruce up the documentation, and to fix the CSV output when the mmm summary is used to reflect that we are now not summarizing across models. I like how you've added the NCR route here as well, that is definitely convenient. This looks good from my perspective. I'll let you merge this, or we can keep it open for a bit if you like. The reason to maybe keep it open is because we'll be adding the IEM Historical here once that coverage is ready.

templates/permafrost/point.html
```diff
@@ -14,6 +14,11 @@ <h3>Geophysical Institute Permafrost Lab (GIPL) model output</h3>
   Query a specific time slice of GIPL model outputs (1 km spatial resolution)
   for ten variables for a single point specified by latitude and longitude.
 </p>
+<p>
+  <strong>NOTE: </strong>The time range for historical data is <strong>1902-2015</strong>
```
charparr, line 18: Thanks for adding this note Bob, I can see where this would be confusing!

routes/permafrost.py
```diff
@@ -92,43 +98,82 @@ def package_obu_vector(obu_vector_resp):
...
-    gipl1km_wcps_point_pkg = {}
+    gipl1km_wcps_point_pkg = dict()
     summary_methods = ["min", "mean", "max"]
-    for resp, stat_type in zip(gipl1km_wcps_resp, summary_methods):
-        gipl1km_wcps_point_pkg[f"gipl1km{stat_type}"] = {}
-        for k, v in zip(gipl1km_dim_encodings["variable"].values(), resp):
-            gipl1km_wcps_point_pkg[f"gipl1km{stat_type}"][k] = round(v, 1)
+    if package_type == "historical":
+        gipl1km_wcps_point_pkg["Historical"] = dict()
+        gipl1km_wcps_point_pkg["Historical"]["Historical"] = dict()
```
charparr, line 151: I'm wondering if one of these dict keys should be "CRU-TS 4.0" to indicate the "model" so-to-speak used for the historical baseline?

charparr (approved):
Hey Bob this looks really good!

I tried quite a few different queries, for both CSV and JSON returns. Everything is working as I'd expect. I also tested the previous endpoint for the 4 km Melvin GIPL + other permafrost layers, and that service is intact. I think the documentation is looking for this endpoint as well.

routes/permafrost.py
BobTorgerson: Yes, you are correct that we should specify the model as CRU-TS 4.0 and the scenario as Historical. Thanks for catching this! I've added that change to the most recent commit.

brucecrevensten (commented):
Points for discussion before merge,

 - can we reduce the number of round-trips to Rasdaman by a bunch (perhaps through removing min/max at this time),
 - chat with team about versioning for the API endpoints plus/and structure of API requests (like, api path structure) for the different flavors of permafrost we have.  I get that "old vs. new" are distinct right now but/and maybe we want to reflect a bit -- there's no right answer and it'd be valid to move ahead with how things are structured now
 - minor quibble with NCR endpoint, consider the main keys being the date ranges instead of numerical indices (look at temp/precip endpoints as example)

cstephen:
This PR has been superseded by PR #266, which was based off of the `gipl_ncr_request` branch with some of its git history dropped, so I'm going to close this PR to avoid confusion.

PR #273, opened 2023-07-17
Adds 500 error image to 500 errors coming from API.

### LINKED ISSUE #76: Customize 500 error page

BobTorgerson:
This adds an image to the 500 errors that come from the API. A way to test this is to add this line to the top of a function in the API that you can call from an endpoint:

return render_template("500/server_error.html"), 500

Closes #76


cstephen (approved):
I confirmed that the new image shows up + looks awesome!

PR #275, opened 2023-07-17
Replace UTF-8 characters in CSV metadata that Excel struggles with

### LINKED ISSUE #148: UTF8 characters in generated CSV metadata render incorrectly in Excel
There's two classes of problems here: (1) UTF8 chars for place names (not sure what to do about this) and (2) characters we use in the metadata (usually dashes or the º symbol).  Excel for Mac (V16) doesn't handle these and displays garbage.

For (2) here's an example:
Screen Shot 2022-03-22 at 7 06 03 PM

The dash characters are just UTF8 chars for the last two (copy/paste thing) and easy to correct.  For the data sources, though, it could be a touch more nuanced because (C) looks a bit like "Copyright" ... so maybe (deg C) or something.  Or perhaps it's just "this is an Excel issue, won't fix."

This works fine in Numbers and LibreOffice.

### LINKED ISSUE #150: Fix dashes/hyphens in permafrost CSV era definitions in metadata section
In the metadata section of the permafrost CSVs, the following eras appear to be using a special dash/hyphen character that the other eras do not have. They show up like this when imported into Excel:

```
# 2075 represents 2061 â€“ 2090
# 2095 represents 2086 â€“ 2100
```

The other eras do not have this problem. We know you need to manually tell Excel to open CSV files with the UTF-8 character set, but the date ranges should all be using consistent characters (one that renders properly in Excel without manual intervention, preferably).

Xref: #148

cstephen:
Closes #148.
Closes #150.

This PR replaces a couple problematic UTF-8 characters (`–` and `°`) in permafrost CSV metadata with unproblematic characters/strings.

To test, download a permafrost CSV. For example:

http://127.0.0.1:5000/permafrost/point/65.0628/-146.1627?format=csv

And try opening it with Excel to confirm that the binary gibberish is gone from the date ranges in these two lines:

```
# 2075 represents 2061 - 2090
# 2095 represents 2086 - 2100
```

And that all `°` characters have been replaced with the string `deg `.

### DIFF
--- a/routes/permafrost.py
+++ b/routes/permafrost.py
@@ -45,8 +45,8 @@ wfs_targets = {
 }
 
 titles = {
-    "gipl": "Melvin et al. (2017) GIPL 2.0 Mean Annual Ground Temperature (°C) and Active Layer Thickness (m) 4 km Model Output",
-    "gipl1km": "GIPL 2.0 Mean Annual Ground Temperature (°C), Permafrost Base, Permafrost Top, and Talik Thickness (m) 1 km Model Output",
+    "gipl": "Melvin et al. (2017) GIPL 2.0 Mean Annual Ground Temperature (deg C) and Active Layer Thickness (m) 4 km Model Output",
+    "gipl1km": "GIPL 2.0 Mean Annual Ground Temperature (deg C), Permafrost Base, Permafrost Top, and Talik Thickness (m) 1 km Model Output",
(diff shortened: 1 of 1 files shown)


BobTorgerson (approved):
The CSVs download for me and I can open them in Excel without issue, I approve these changes to the permafrost CSVs.

PR #277, opened 2023-07-19
Sea Ice improvement

### LINKED ISSUE #246: Sea Ice endpoint: Don't return values larger than 100 as this is a percentage of ice
For some locations, there are values of sea ice that are showing as 253 for percentage of ice coverage. I believe that along with 120, 254, and 255, the value 253 represents a "no data" value indicating something from the data that does not need to be shown in our output. 

Figure out where the documentation regarding those values resides on this website: [NSIDC Site](https://nsidc.org/data/g10010/versions/2) and set the 253 value to nodata if found in the documentation.

BobTorgerson:
Adds a check for any sea ice concentration above 100% and provides an additional error code of 253 to the list of known no data values.

Closes #246

### DIFF
--- a/routes/seaice.py
+++ b/routes/seaice.py
@@ -73,8 +73,13 @@ def package_seaice_data(seaice_resp):
     """
     # initialize the output dict
     di = dict()
+
+    # For each year and month, checks to see if sea ice percentage is greater than 100,
+    # and sets the value of the percentage for each month to the returned value or 0.
     for i in range(len(seaice_resp)):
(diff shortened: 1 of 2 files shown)


cstephen (approved):
I found a few example URLs where the 253% problem occurred and confirmed that this PR fixed them:

http://127.0.0.1:5000/seaice/point/65.47/-166.97/
http://127.0.0.1:5000/seaice/point/66.15/-163.73/
http://127.0.0.1:5000/seaice/point/68.52/-166.20/

0 now appears where 253 appeared before, so all looks good!

PR #278, opened 2023-07-20
Updates CSV filename based on month if provided in URL.

### LINKED ISSUE #252: Add month (January or July) to CSV metadata or column for temperature min/mean/max endpoint
If you download the CSV file from this URL for example (taken from one of the CSV download buttons in Arctic EDS):

https://earthmaps.io/mmm/temperature/july/all/60.1043/-149.442?format=csv&community=AK354

If you then look at the CSV file, there is no indication that the CSV data is for a particular month, or what month (July vs. January). It appears to be using the same set of metadata and columns as the annual data. So, we just need to fix up the CSV files for temperature `mmm` (min/mean/max) endpoints to be more clear.

BobTorgerson:
This PR adds the ability to see the month selected in the file name when that month was included as part of the API endpoint URL.

Example: http://localhost:5000/mmm/temperature/july/all/60.1043/-149.442?community=AK354&format=csv

Try a few locations, some that are lat / lon coordinates and some that are communities.

Closes #252

### DIFF
--- a/routes/taspr.py
+++ b/routes/taspr.py
@@ -851,19 +851,28 @@ def return_csv(csv_data, var_ep, place_id, lat=None, lon=None, month=None):
 
     place_name, place_type = place_name_and_type(place_id)
 
-    if place_name is not None:
+    if month is not None:
+        if place_name is not None:
+            filename = (
+                    mmm_dim_encodings["months"][month]
+                    + " "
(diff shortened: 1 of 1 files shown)


cstephen (approved):
I tried a few different lat/lon and community locations and this is working as expected 👍

PR #281, opened 2023-07-20
Silences warnings in the console window.

### LINKED ISSUE #128: Handle some spurious (?) warnings?
Two warnings show up in the stdout --

```
"UserWarning: The Shapely GEOS version (3.10.2-CAPI-1.16.0) is incompatible with the GEOS version PyGEOS was compiled with (3.10.1-CAPI-1.16.0). Conversions between both will be slow."```

```ShapelyDeprecationWarning: The array interface is deprecated and will no longer work in Shapely 2.0. Convert the '.coords' to a numpy array instead.```

We are not convinced either of these are legit warnings - they may be a consequence of the ongoing Shapely development toward a 2.0 with tighter pyGEOS integration (see https://shapely.readthedocs.io/en/stable/migration.html). We could downgrade Shapely to 1.7 (unknown if/what impacts for other packages), shush the warnings, or leave them in place to address at a later time.

BobTorgerson:
This PR silences all Python warnings in the console window. To test, check your console and make sure that none of the warnings are noted.

Closes #128

### DIFF
--- a/application.py
+++ b/application.py
@@ -1,6 +1,10 @@
 from datetime import datetime
 from flask import Flask, render_template
 from flask_cors import CORS
+import warnings
+
+# Silence warnings in console before importing routes
+warnings.filterwarnings("ignore")
 
 from routes import *
 


brucecrevensten:
I think the goal of #128 was to address the warnings, not silence them?  Not sure why we'd want to hide warnings in general.

BobTorgerson:
It specifically said in the issue "shush the warnings" as an option, but I'll go ahead and close this PR.

PR #282, opened 2023-07-20
Uses invalid_year template for degree_days /mmm route

### LINKED ISSUE #191: Use `422/invalid_year.html` for invalid years in other parts of `/mmm` endpoint (namely temp, precip)
The `422/invalid_year.html` template was introduced for the degree days queries developed in #188 . Make use of this in other parts of this endpoint. This might qualify as a sub-ticket of #97 .

BobTorgerson:
If min and max years are given for /mmm/degree_days endpoint, it will give a 422 error and use the invalid_year template.

To test this PR, try this endpoint:

http://localhost:5000/mmm/degree_days/freezing/hp/60/-156/1800/2099

Closes #191

### DIFF
--- a/routes/degree_days.py
+++ b/routes/degree_days.py
@@ -13,11 +13,7 @@ from validate_request import (
     validate_latlon,
     project_latlon,
 )
-from validate_data import (
-    nullify_and_prune,
-    postprocess,
-    place_name_and_type
-)
+from validate_data import nullify_and_prune, postprocess, place_name_and_type
 from config import WEST_BBOX, EAST_BBOX
 from . import routes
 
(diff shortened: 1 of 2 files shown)


cstephen (changes requested):
This is pretty close to being what we want, but I noticed that when the error page says:

> Valid years for this endpoint are _min_year_ - _max_year_

The min and max years it shows are the invalid years provided by the user, not the acceptable min/max years.

The acceptable min/max years are stored in the `years_lu` dictionary, but because the acceptable year range depends on whether the user is requesting historical, projected, or "horp", we'll need some additional logic to decide what the correct min/max year range is depending on what the user is requesting.

cstephen (approved):
I've tried several combinations of degree day endpoints and year ranges and this now works as expected. The displayed min/max years adapt to whether the user is asking for historical, projected, or "hp". All looks well!

Btw, I had to change the test URL in the PR description from:

http://localhost:5000/mmm/degree_days/freezing/hp/60/-156/1800/2099

To:

http://localhost:5000/mmm/degree_days/freezing_index/hp/60/-156/1800/2099

But it works as intended after that.

PR #283, opened 2023-07-21
Adds cache control header fields to keep returned data for 90 days

### LINKED ISSUE #73: Set cache-control headers
We want Flask to decorate all request responses with some standard cache headers.

```
Cache-Control: max-age=7776000, s-maxage=7776000, must-revalidate, proxy-revalidate, immutable
```

^ Along those lines I think.  Bob/Craig should review!  Ideally this is a decorator that can handle every request, not a per-route situation.

BobTorgerson:
This PR adds the cache-control headers to our site which can inform upstream caches of how long we would like to cache a file for future reference. When pointed at the Varnish cache for either Rasdaman (maps.earthmaps.io) or the data API (earthmaps.io), it is specifying a max age of 90 days and matches the length of cached time on the Varnish servers.

To test this, you can go to the updated development.earthmaps.io site and look at the returned headers for max age in the cache-control section of the response header: http://development.earthmaps.io/alfresco/veg_type/area/19080309

Closes #73

### DIFF
--- a/fetch_data.py
+++ b/fetch_data.py
@@ -93,7 +93,8 @@ async def make_get_request(url, session):
     Returns:
         Query result, deocded differently depending on encoding argument.
     """
-    resp = await session.request(method="GET", url=url)
+    cache_header = {"Cache-Control": 'max-age=7776000, s-maxage=7776000'}
+    resp = await session.request(method="GET", url=url, headers=cache_header)
     resp.raise_for_status()
 
     # way of auto-detecting encoding from URL
--- a/fetch_data.py
+++ b/fetch_data.py
@@ -93,7 +93,7 @@ async def make_get_request(url, session):
     Returns:
         Query result, deocded differently depending on encoding argument.
     """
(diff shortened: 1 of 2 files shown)


fetch_data.py
```diff
@@ -93,7 +93,8 @@ async def make_get_request(url, session):
     Returns:
         Query result, deocded differently depending on encoding argument.
     """
-    resp = await session.request(method="GET", url=url)
+    cache_header = {"Cache-Control": 'max-age=7776000, s-maxage=7776000'}
```
cstephen, line 96: Should `s-maxage=7776000` be removed from here, too? I don't know one way or the other, just curious!

cstephen (changes requested):
I've confirmed that the `max-age` cache control header shows up on development.earthmaps.io, and locally too. I added one comment to the code review. Please confirm one way or another if this should be fixed, then this looks good to merge. Thanks!

BobTorgerson:
@cstephen Go ahead and merge whenever you're ready!

cstephen (approved):

PR #284, opened 2023-07-24
Removes references to the CORDEX indicators in the API

### LINKED ISSUE #268: Replace references to CORDEX in indicators endpoints
There are some variables and code comments that reference CORDEX that should be removed or replaced with the name of the dataset we ended up using.

BobTorgerson:
This PR removes references and functionality that was utilized for the CORDEX indicators that will no longer be used. 

Closes #268


cstephen (approved):
I've confirmed that all references to CORDEX have been removed from the code, and that point and area queries still work against the "base" indicators endpoint.

PR #286, opened 2023-07-26
Consolidate and standardize disparate CSV implementations into common set of functions

### LINKED ISSUE #98: Ensure that shared general CSV code is used by all applicable endpoints
The new ALFRESCO PR #94 introduces some shared code for generalizing the CSV export functionality. Evaluate the feasibility of applying this to all applicable endpoints and ensure that the CSV export is the consistent across them, and only carry this out **if it makes sense to do so**. This might require modification of the shared code to make it more generalized. It could be that some or all endpoints won't be able to make use of this shared function anyway without significant modification.

image

### LINKED ISSUE #147: Consolidate CSV export code in taspr, permafrost, and alfresco endpoints
There is a lot of similar code in the CSV functionality in the taspr, permafrost, and alfresco endpoints. Much of this code can probably be modified & consolidated. Look at the `create_csv` and `return_csv` functions for example.

cstephen:
Closes #98.
Closes #147.

The primary goal of this PR is to consolidate all CSV export code into a shared set of functions. Before this PR, most endpoints with CSV support implemented their own versions of `create_csv` and associated functions and tended to be inconsistent with their approaches, making them hard to maintain on the whole.

This created situations where, for example, some CSVs had location metadata, and others did not, because each endpoint handled metadata differently. This PR fixes the issue by removing all CSV functions from route/endpoint files and creating a new `csv_functions.py` file that handles all aspects of CSV processing across all endpoints.

With the new generalized `create_csv` function inside `csv_functions.py` comes a more standardized approach to CSV processing. In all cases, endpoint data is passed into the `create_csv` function with the same structure/hierarchy as the corresponding JSON output for the endpoint. Then, any preprocessing needed to make the data CSV-ready, such as adding a more granular nesting structure to the dict, is handled in an endpoint-specific function (e.g., `taspr_csv`) within `csv_functions.py`.

This PR also moves all data postprocessing functions like `postprocess`, `nullify_nodata`, and `nullify_and_prune` into a new `postprocessing.py` file. This was done because I kept running into circular import errors while working on this PR, and pulling the postprocessing functions out of `validate_data.py` seemed like the most sane way to solve the issue, and also just seems like a good move in general since data validation and postprocessing are fairly distinct from each other.

To test, I've created the following spreadsheet of all available CSV endpoints with links to the development.earthmaps.io vs. local URLs for each:

https://docs.google.com/spreadsheets/d/14-Bzdhy1x8vpEwFtc32qL54hKK4Usmta_Ukj4WL5nus/edit?usp=sharing

This makes it easy to:

- Confirm that all CSV endpoints still work
- See the differences between old vs. new CSVs

In some cases, the structure of the CSVs has changed drastically to make them more "spreadsheety". For example, check out the old GIPL CSV:

https://development.earthmaps.io/permafrost/point/gipl/63.0628/-146.1627?format=csv

Compare that with the new version of the CSV:

http://localhost:5000/permafrost/point/gipl/63.0628/-146.1627?format=csv

Instead of creating a separate row for each model/year/scenario/variable combination, the new CSV above creates a single row for model/year/scenario and then adds all of the corresponding variables to their own columns on the same row, reducing the file size by ~80%. This same approach has been applied to all other endpoint CSVs that listed each variable on its own line, making the CSV files more compact overall.


BobTorgerson (approved):
This is a really great and much needed improvement to consolidating the CSV generation code! Quite a large undertaking, thanks so much for the improvements! I approve and will merge this code into main for us to use for future endpoints.

PR #306, opened 2023-08-31
Use lxml instead of built-in ElementTree module or string splitting tricks to parse coverage XML

### LINKED ISSUE #243: Use XPath / `lxml` to parse DescribeCoverage
Functions like this are used to parse DescribeCoverage responses and create the skeleton of the data package that ultimately is filled in by a GetCoverage request or similar:

```python
def get_xml_str_between_tags(meta_xml_str, tag, occurrence=1):
    """Get string encapsulated by a known XML tag. Use this function to retrieve time axis values that are not encapsulated by a dictionary and/or are not within the metadata 'Encoding' block.
    Arguments:
        meta_xml_str (str): string representation of the byte XML response from the WCS DescribeCoverage request
        tag (str): the xml string that encapsulates the desired information, e.g., 'gmlrgrid:coefficients'
        occurrence (int): the occurrence of the tag to parse. some tags are repeated several times in the XML response
    Returns:
        str_within_tag (str): string encapsulated by the provided XML tag
    """
    tag_open = f"<{tag}>"
    tag_close = f"</{tag}>"
    str_after_tag = meta_xml_str.split(tag_open)[occurrence]
    str_within_tag = str_after_tag.split(tag_close)[0]
    return str_within_tag
```

This is kind of hacky and probably not the most flexible way to do thist. We should extract these expressions using XPath.

cstephen:
Closes #243.

This PR updates (and renames) the `get_xml_str_between_tags` function to use the lxml module instead of string splitting tricks to parse coverage XML from Rasdaman. It also makes a small tweak to the `parse_meta_xml_str` function to use lxml instead of Python's built-in ElementTree module so we're not using two different XML parsing modules in the same file.

I originally tried to use the built-in ElementTree module for `get_xml_str_between_tags` as well, rather than using the third-party lxml module, but found that lxml makes parsing namespaced elements _much_ easier.

To test, run `pipenv install` to make sure the lxml module gets installed. Then verify that everything worked just like before. If there are any errors with XML parsing, these will typically show up in the terminal when the API starts up. But double check that a handful of endpoints are returning data as well.


brucecrevensten (approved):
I did a `pipenv clean` and a `pipenv install` then re-ran and it worked!  I tried a few various endpoints, nothing broke.  Unsure if I exercised the code enough, but willing to merge.

PR #324, opened 2023-09-13
Modifies documentation to use academic citation format

brucecrevensten:
Simplifies the Data Sources section a bit further, reducing the number of links that we may need to maintain & simply referring out to certain external works in general instead of specific files (more brittle).

Also some small changes elsewhere in the documentation.

To test this:

 - Visit each documentation endpoint.  Each should have a citation-style reference out to the relevant academic work(s) [[ some won't -- the endpoints we are removing, or ones we don't have final academic references for yet -- Landfast Ice, Design Indices, Beetles, Glaciers, Mean Annual Ground Temp ]]
 - Check that the table isn't mangled and the layout is "reasonable" (we don't super-care about looks here)


cstephen (changes requested):
Proper academic citations are a big improvement, and this mostly looks good. I found one important thing that needs changed:

- [ ] Jeremy's last name is missing an `l` in the Snowfall Equivalent citation.

And a few nitpicky findings that may or may not have been intentional. Feel free to ignore if intentional or too nitpicky!

- [ ] The source data section for Geology is not in table format.
- [x] The Elevation endpoint says "Citation" instead of "Academic reference" in the column headers, unlike the other tables. The Wildfire endpoint says "Citation / Academic reference".
- [x] Some of the tables have a horizontal line across the bottom of the table (which is added by the `tfoot` tag), and some don't. We may want to decide on one style or the other.

brucecrevensten:
The nitpicky bits were intentional, just because they were a better fit (imho) for those specific endpoints.

PR #325, opened 2023-09-13
Remove historical/projected URL paths, rename remaining /mmm endpoints, and control summarization via "?summarize=mmm" GET parameter

### LINKED ISSUE #292: Remove `mmm` / date-range summaries
After team discussion, we probably want to remove the `/mmm` summarization.  We need to review this on a per-variable basis first, though, and consider the front-end integration implications before implementing.

Related, we probably will need to remove any date-range-summaries as well.

### LINKED ISSUE #315: Remove `/historical` and `/projected` and `/hp` and just serve them all together
We don't have a concrete use case for separating out the historical/projected (and it varies by dataset).  Serve the whole time series without further temporal clarification.

cstephen:
Closes #292.
Closes #315.

After discussing our game plan in a Slack huddle yesterday, this PR:

- Removes "/mmm" from the beginning of all "/mmm" endpoint paths.
- Removes "historical" and "projected" as options from former /mmm endpoint paths so that each endpoint always returns both historical and projected data.
- Removes the "hp" option from former /mmm endpoint paths but replaces them with functionally identical endpoints by appending `?summarize=mmm` to the end of URLs. For example, `/degree_days/heating/65.0628/-146.1627?summarize=mmm` now serves the same data that `/mmm/degree_days/heating/hp/65.0628/-146.1627` did.
- Removes the "all" option from former /mmm endpoints and replaces them with functionally identical endpoints without any horp parameter. For example, `http://localhost:5000/degree_days/heating/65.0628/-146.1627` now serves the same data that `/mmm/degree_days/heating/all/65.0628/-146.1627` did.

To test, I've put together this Google Sheet of all former `/mmm` endpoints and their replacement endpoints where applicable:

https://docs.google.com/spreadsheets/d/1gV6p60UmOKzJW71DFOW6DPx1g0_vO5wIYAQgmqhI4Jg/edit#gid=0

For one final mega-test, verify that the `/eds/all` endpoint (e.g., http://localhost:5000/eds/all/65.0628/-146.1627) returns all of the same data that it did before, with two caveats:

- This branch no longer returns any output for the `design_freezing` and `design_thawing` components because we are planning to remove these soon anyway, and it would have taken extra work to keep them working in this branch.
- Sometimes the `ddmean` values returned from the `/eds/all` endpoint will differ by a value of 1 vs. the `main` branch due to rounding differences. The code to calculate averages across years for the Arctic EDS has changed quite a bit because it had relied fairly heavily on the historical/projected options that have been removed, and this is what introduced the rounding difference.

Because this PR retains the functionality of the /mmm endpoints and simply makes them accessible in a different way (`?summarize=mmm`), I did not remove the ability to summarize across year ranges since this is almost certainly still useful.


brucecrevensten (changes requested):
See the testing spreadsheet, highlighted routes in red aren't working.

I don't think it's specific to this branch with main merged in (i.e. I don't think it's a merge issue), it looked like it was happening even when I walked the commit tree back ~15+ commits.

This is the code that appears to be failing:
`dd_dim_encodings = asyncio.run(get_dim_encodings("heating_degree_days"))`

PR #329, opened 2023-09-17
Updates footer, slight style changes

### LINKED ISSUE #305: Update footer with most current legal phrasing
xref https://github.com/ua-snap/arctic-eds/issues/293

brucecrevensten:
Closes #305 

Test by:

 - Confirm that the language of the footer for AA/EO matches exactly `UA is an affirmative action/equal opportunity employer, educational institution and provider and prohibits illegal discrimination against any individual: http://www.alaska.edu/nondiscrimination` (we won't deviate from this anywhere)
 - notice the slightly gentler colors for the background.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -1,12 +1,19 @@
+body {
+	background-color: #f3f3ed;
+	color: #061728 !important;
+}
+
 .headerbanner {
 	padding: .5rem;
 	text-transform: uppercase;
 	text-align:  center;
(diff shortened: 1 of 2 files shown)


cstephen (changes requested):
This looks good, except the `http://` is missing from the link AA/EO link text (not sure how much this matters), but also the link is broken because `http://` is missing from the link as well.

PR #334, opened 2023-09-19
del 4km GIPL; del unused luts/imports

### LINKED ISSUE #297: Remove GIPL1 (Melvin) from API / backend storage
Double-check all documentation when this is done.  We should only refer to GIPL2.0.

Joshdpaul:
This PR closes #297 by removing all references to the old 4km GIPL data. 

The `run_point_fetch_all_permafrost` function called by the `@routes.route("/permafrost/point/<lat>/<lon>")` endpoint now packages the new 1km GIPL data instead (see line 256 in `permafrost.py`).

To test, run the `flask` application as described in `README.md`, and try:

[http://127.0.0.1:5001/permafrost/point/64/-145](http://127.0.0.1:5001/permafrost/point/64/-145)

Try out some other coordinates to make sure you are seeing data (or errors) as expected. You should see the first point data package titled "gipl_1km" instead of "gipl", which indicates the return from the `package_gipl1km_point_data` function at line 263 in `permafrost.py`. 

**Note:** The default endpoint link from the permafrost template ([/permafrost/point/65.0628/-146.1627](http://127.0.0.1:5001/permafrost/point/65.0628/-146.1627)) does not return the "gipl_1km" point data...this return must be cached somehow? Any suggestions on how to fix that would be welcome!

### DIFF
--- a/routes/permafrost.py
+++ b/routes/permafrost.py
@@ -19,10 +19,8 @@ from validate_request import (
 from postprocessing import nullify_and_prune, nullify_nodata, postprocess
 from . import routes
 from config import GS_BASE_URL, WEST_BBOX, EAST_BBOX
-from luts import permafrost_encodings  # for the Melvin 4 km (NCR) data
 
-# rasdaman coverages
-permafrost_coverage_id = "iem_gipl_magt_alt_4km"
+# rasdaman coverage
 gipl_1km_coverage_id = "crrel_gipl_outputs"
 
 gipl1km_dim_encodings = asyncio.run(
@@ -41,8 +39,7 @@ wfs_targets = {
 }
 
 titles = {
-    "gipl": "Melvin et al. (2017) GIPL 2.0 Mean Annual Ground Temperature (deg C) and Active Layer Thickness (m) 4 km Model Output",
-    "gipl1km": "GIPL 2.0 Mean Annual Ground Temperature (deg C), Permafrost Base, Permafrost Top, and Talik Thickness (m) 1 km Model Output",
+    "gipl_1km": "GIPL 2.0 Mean Annual Ground Temperature (deg C), Permafrost Base, Permafrost Top, and Talik Thickness (m) 1 km Model Output",
     "jorg": "Jorgenson et al. (2008) Permafrost Extent and Ground Ice Volume",
     "obupfx": "Obu et al. (2018) Permafrost Extent",
 }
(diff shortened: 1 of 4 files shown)


cstephen (changes requested):
This looks great overall and getting rid of the old GIPL data is a huge step in the right direction! I found a few things to fix that are a little bit tangential to this PR, but important nonetheless:

- [x] The CSV download functionality no longer works for the  /permafrost/point endpoint (e..g, `/permafrost/point/64/-145?format=csv`). I think all you'll need to do to fix this is update the coords/values here to reflect the new JSON keys and values for GIPL 2.0: https://github.com/ua-snap/data-api/blob/e785c40d3c3b9f25b05f75a2c21cba08633872b3/csv_functions.py#L401-L409
- [x] We'll also need to update the permafrost CSV metadata lines to remove the mention of the "alt" variable (active layer thickness) since that was available in GIPL 1.0 but not GIPL 2.0. Also since the GIPL 2.0 data is not served in pre-summarized eras, we'll want to remove all CSV metadata lines like this too: "2025 represents 2011 - 2040". See here: https://github.com/ua-snap/data-api/blob/e785c40d3c3b9f25b05f75a2c21cba08633872b3/csv_functions.py#L415-L422
- [ ] The documentation here needs to be updated to describe the new GIPL output structure and variables served from the /permafrost/point endpoint. https://github.com/ua-snap/data-api/blob/e785c40d3c3b9f25b05f75a2c21cba08633872b3/templates/documentation/permafrost.html#L212-L224 https://github.com/ua-snap/data-api/blob/e785c40d3c3b9f25b05f75a2c21cba08633872b3/templates/documentation/permafrost.html#L246-L260

Joshdpaul:
OK yep I totally glossed over the CSV export bit here!

I think I have addressed all 3 requested changes above with [70f0095](https://github.com/ua-snap/data-api/pull/334/commits/70f0095384398c94012af3918ca39bd237af57c0) and [edc91fe](https://github.com/ua-snap/data-api/pull/334/commits/edc91fec4ecabc5b195a6176a67f977a8ad21398) ... The changes I made to the HTML doc look OK to me but might require a detailed look from someone more familiar with the nested JSON formatting!

cstephen (approved):
I've verified that the CSV export functionality works and the documentation and CSV metadata have been fixed. Looks good!

routes/permafrost.py
charparr, line 1: I think you can delete one more function!

This one is definitely a relict of the Melvin dataset:

```python
def combine_gipl_poly_var_pkgs(magt_di, alt_di):
    combined_gipl_di = {}
    for era in magt_di.keys():
        combined_gipl_di[era] = {}
        for model in magt_di[era].keys():
            combined_gipl_di[era][model] = {}
            for scenario in magt_di[era][model].keys():
                combined_gipl_di[era][model][scenario] = {}
                combined_gipl_di[era][model][scenario]["magt"] = magt_di[era][model][
                    scenario
                ]
                combined_gipl_di[era][model][scenario]["alt"] = alt_di[era][model][
                    scenario
                ]
                combined_gipl_di[era][model][scenario]["statistic"] = "Zonal Mean"
    return combined_gipl_di
```

charparr (changes requested):
Awesome! Endpoints are working as I'd expect and the documentation looks good. I think you can remove one more function, then this should be good to merge.

Issue #339, opened 2023-09-21
Consider a API-wide date range validation function for rasdaman coverages


Joshdpaul:
The rasdaman coverage metadata would be retrieved in the `routes/*.py` modules, along with start/end years, and then passed to this function.  The goal is to make it useable for all rasdaman coverages to avoid performing this validation inside of every `routes/*.py` module. 

The validation function would accept endpoint start/end dates and a coverage-specific date index as arguments. Returns True/False validity of endpoint start/end dates.

cstephen:
Xref: https://github.com/ua-snap/data-api/issues/617

This issue is no longer relevant since we're going to remove start/end year functionality from all endpoints.

Issue #338, opened 2023-09-21
Include date range in CSV export metadata header


Joshdpaul:
If using `?format=csv` argument to export summarized data from an endpoint, include the year range somewhere in the CSV file (eg, as metadata in the header, in the filename of the CSV itself, etc.)

Make sure to include the cases where no year range was specified in the endpoint itself, where summaries are computed on the entire date range.

Issue #337, opened 2023-09-20
Bad ?arg=value requests still give data


Joshdpaul:
For example, see the `fetch_mmm_point_data()` function in  `taspr.py`. The script checks for a `?summarize=mmm` argument via `if request.args.get("summarize") == "mmm":` condition. However any typo in this argument (or any argument at all for that matter) will default this condition to return 'None' in which case the script moves to the `else:` condition and continues to execute.

For example, the endpoints below with bad arguments all return data, but should probably return an error instead to avoid confusion:

http://127.0.0.1:5001/temperature/65.0628/-146.1627/1940/2060?summarize=max
http://127.0.0.1:5001/temperature/65.0628/-146.1627/1940/2060?summarize=mmmm
http://127.0.0.1:5001/temperature/65.0628/-146.1627/1940/2060?summary=mmm
http://127.0.0.1:5001/temperature/65.0628/-146.1627/1940/2060?foo=bar

Consider using a zero-length argument dictionary check first, eg:

`if len(request.args) == 0:  
     return "default endpoint"`

then check for arguments explicitly, eg: 

`elif: "summarize" in request.args:  
        if request.args.get("summarize") == "mmm":  
                 summarize the data  
        else: throw an error`

Joshdpaul:
This issue was addressed in  #341 but only for the `permafrost.py` route. Consider explicit argument validation: either route-specific (as done in `permafrost.py` or as new API-wide function (see #339 ).

brucecrevensten:
https://marshmallow.readthedocs.io/en/stable/

Joshdpaul:
Affected endpoints:

https://earthmaps.io/degree_days/
https://earthmaps.io/hydrology/
https://earthmaps.io/snow/
https://earthmaps.io/taspr/

PR #341, opened 2023-09-21
Revise permafrost mmm

Joshdpaul:
This PR:

- removes the `<summarize>` endpoint in `permafrost.py` and replaces it with a request argument `?summarize=mmm`. The `summarize` argument is left unchanged in the `fetch_gipl…()`, `run_fetch_gipl…()` functions, so the `run_ncr_requests()` function should work as usual.
- Adds explicit validation of any `?summarize=` and `?format=` args. Previously, bad arguments would default to None which continued to produce data (eg, `?summarize=mean` would still produce a non-summarized JSON result, which was confusing). This addresses issue #337 but only for the `permafrost.py` route. Explicit validation like this, if repeated in every `route/*.py` module, could become very redundant and could perhaps be moved to its own module specifically for validating request arguments (see issue #339). 
- Adds a default to summarize all years in the dataset if a date range is not provided in the endpoint but `?summarize=mmm` is present. For example, `/permafrost/point/gipl/62/-145?summarize=mmm` will now return min/mean/max summaries across all years in the coverage. Previously, this request would produce a non-summarized JSON result, which was confusing. This leads directly to issue #338, which would provide more information in the CSV output so the user knows what dates have been summarized, which is especially important for implicit cases like this.
- Updates the template documentation `permafrost.html` by removing the MMM section and including a note about the `?summarize=mmm` argument. JSON structure examples at the bottom of the template were left unchanged.

To test, run the `flask` application as described in `README.md`, and try:

Standard endpoints like:

	/permafrost/point/gipl/62/-145
	/permafrost/point/gipl/62/-145/2030/2040
	/permafrost/point/62/-145

GIPL endpoints with `?summarize=mmm` argument, `?format=csv` argument, and combined `?summarize=mmm&format=csv` arguments:

	/permafrost/point/gipl/62/-145?summarize=mmm
	/permafrost/point/gipl/62/-145/2030/2040?summarize=mmm	
	/permafrost/point/gipl/62/-145?format=csv
	/permafrost/point/gipl/62/-145/2030/2040?format=csv
	/permafrost/point/gipl/62/-145?summarize=mmm&format=csv
	/permafrost/point/gipl/62/-145/2030/2040?summarize=mmm&format=csv

Bogus arguments to make sure they don’t return any data:

	/permafrost/point/gipl/62/-145?foo=bar
	/permafrost/point/gipl/62/-145?summarize=mean
	/permafrost/point/gipl/62/-145/2030/2040?summarize=mean
	/permafrost/point/gipl/62/-145?format=xlsx
	/permafrost/point/gipl/62/-145/2030/2040?format=xlsx
	/permafrost/point/gipl/62/-145?summarize=mean&format=csv
	/permafrost/point/gipl/62/-145/2030/2040?summarize=mmm&format=xlsx

### DIFF
--- a/templates/documentation/permafrost.html
+++ b/templates/documentation/permafrost.html
@@ -1,70 +1,37 @@
-{% extends 'base.html' %}
-{% block content %}
+{% extends 'base.html' %} {% block content %}
 <h2>Permafrost Data</h2>
 
 <p>
-  These endpoints provide access to permafrost information for Alaska, including outputs from academic research on
-  permafrost type and extent as well as ground temperatures.
+  These endpoints provide access to permafrost information for Alaska, including
+  outputs from academic research on permafrost type and extent as well as ground
+  temperatures.
 </p>
 
 <p>
-  Annual projections of permafrost top and base depths, talik thickness, and mean annual ground temperature at seven
-  different depths were provided by the GIPL 2.0 model at a resolution of 1km for years 2021&ndash;2120. These
-  projections are provided for GFDL-CM3 and NCAR-CCSM4 model outputs, as well as a 5-model average, under the RCP
-  4.5 and RCP 8.5 emissions scenarios.
+  Annual projections of permafrost top and base depths, talik thickness, and
+  mean annual ground temperature at seven different depths were provided by the
+  GIPL 2.0 model at a resolution of 1km for years 2021&ndash;2120. These
+  projections are provided for GFDL-CM3 and NCAR-CCSM4 model outputs, as well as
+  a 5-model average, under the RCP 4.5 and RCP 8.5 emissions scenarios.
 </p>
 
 <p>
-  Permafrost extent and ground ice volume data from 2008 were provided by Jorgenson et al. Mean annual top of permafrost
-  ground temperature and permafrost extent data from 2000&ndash;2016 were provided by Obu et al.
+  Permafrost extent and ground ice volume data from 2008 were provided by
+  Jorgenson et al. Mean annual top of permafrost ground temperature and
+  permafrost extent data from 2000&ndash;2016 were provided by Obu et al.
 </p>
 
-<p>Links to academic references and source datasets are included at the bottom of this page.</p>
+<p>
+  Links to academic references and source datasets are included at the bottom of
+  this page.
+</p>
 
 <h3>Service endpoints</h3>
 
 <h4>Geophysical Institute Permafrost Lab (GIPL) model output (point query)</h4>
 
 <p>
-  Query GIPL 2.0 model outputs (1 km spatial resolution) for ten variables for a single point specified by latitude and
-  longitude.
-</p>
-
-<table class="endpoints">
-  <thead>
-    <tr>
-      <th class="endpoint-label">Endpoint</th>
-      <th class="endpoint-url">Example URL</th>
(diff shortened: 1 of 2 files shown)


templates/documentation/permafrost.html
```diff
@@ -84,7 +88,7 @@ <h4>Geophysical Institute Permafrost Lab (GIPL) min/mean/max summaries (point qu
       <td colspan="2">CSV output is also available by appending <code>?format=csv</code> to the URL.</td>
     </tr>
   </tfoot>
-</table>
+</table> -->
```
charparr, line 91: Note sure what the `-->` here is, but I don't know enough about HTML perhaps. Craig will probably have a better idea, but maybe just double check this isn't an editor / linting thing. Ditto with the `!--` on L64. The documentation page is rendering just great though!

templates/documentation/permafrost.html
```diff
@@ -52,12 +52,16 @@ <h4>Geophysical Institute Permafrost Lab (GIPL) model output (point query)</h4>
   </tbody>
   <tfoot>
     <tr>
-      <td colspan="2">CSV output is also available by appending <code>?format=csv</code> to the URL.</td>
+      <td colspan="2">Min/mean/max summaries of projected years are available by appending
```
charparr, line 55: Consider indicating how to chain these args like `/permafrost/point/gipl/63.0628/-146.1627?summary=mmm&format=csv`

charparr (approved):
Hey this is working as I'd expect! I'm able to get summarized and non-summarized data in both JSON and CSV format as I'd expect. I tested a variety of routes, with and without supplying the year range and all worked as expected. Chaining the format and summarize arguments also is working as I'd expect, like so:
`/permafrost/point/gipl/64/-145/2030/2040?format=csv&summarize=mmm`
and the equivalent
`http://127.0.0.1:5000/permafrost/point/gipl/64/-145/2030/2040?summarize=mmm&format=csv`

templates/documentation/permafrost.html
BobTorgerson: These lines should be removed as these are HTML comments: https://github.com/ua-snap/data-api/blob/b8f5a70a6de07f4babd57d25ba801c08e4e54c80/templates/documentation/permafrost.html#L64-L91

templates/documentation/permafrost.html
Joshdpaul: Good idea Charlie! Added with [eef8f61](https://github.com/ua-snap/data-api/pull/341/commits/eef8f613dab5e43f2bb4ff2804a4174824d6bb3f)

cstephen (approved):
I kicked the tires on this a bit and everything is working just as described:

- Chaining `format=csv` with `summarize=mmm` with or without start/end years works as expected
- Bad `summarize` values are caught and an error is returned
- The output & values of http://localhost:5000/permafrost/point/gipl/63.0628/-146.1627?summarize=mmm (without start/end years) exactly matches the output & values of http://development.earthmaps.io/permafrost/point/gipl/63.0628/-146.1627/2021/2120/mmm (i.e., the full set of years) 
- The documentation has been updated to reflect the new endpoint behavior

Awesome work, @Joshdpaul! The GET parameter validation and chaining is also a step up from our other mmm endpoints, so good thinking on those.

Issue #312, opened 2023-09-08
Remove point query for ALFRESCO, enrich documentation


brucecrevensten:
It's not statistically meaningful, so remove the `/point` query here.

Add text to the effect of, "due to the statistical nature of the model that produces these data, it's not meaningful to select a single pixel for output.  This API offers a corresponding Hydrologic Unit (Level 12), or HUC-12, as a reasonable compromise between local specificity and statistically-valid interpretation."

PR #345, opened 2023-09-22
Remove ALFRESCO point data query

Joshdpaul:
This PR closes issue #312 , by removing point data endpoint and functions from `alfresco.py` and references to the point data endpoint in `alfresco.html`. Additional descriptive text explaining the lack of point data endpoints was added to the header description in `alfresco.html`. 

To test, run the `flask` application as described in `README.md`, and verify that the ALFRESCO point data endpoint is no longer referenced on the ALFRESCO documentation page and that a previously existing ALFRESCO endpoint (eg http://127.0.0.1:5001/alfresco/flammability/65/-145) now returns no data.

### DIFF
--- a/routes/alfresco.py
+++ b/routes/alfresco.py
@@ -15,7 +15,6 @@ from fetch_data import *
 from csv_functions import create_csv
 from validate_request import (
     validate_latlon,
-    project_latlon,
     validate_var_id,
 )
 from postprocessing import nullify_and_prune, postprocess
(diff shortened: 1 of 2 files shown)


routes/alfresco.py
charparr, line 1: I think you can safely remove this route synonym too (doesn't matter much, just points to the about page anyway) but I think the fewer synonyms the better, as a general rule.

Line 251: `@routes.route("/alfresco/point/")`

charparr (approved):
Hey Josh, this looks good! I tried a `point` request and got a bad request as expected. All other routes, i.e. `local` are intact. See my comment about the route synonym, delete if ya want, or not, and then I'd say go ahead and merge!

PR #358, opened 2023-10-02
Add hydrology endpoint

Joshdpaul:
This PR adds the basic `/hydrology/point/<lat>/<lon>` endpoint, which includes a point data query with optional `?format=csv` and `?summarize=mmm` arguments. 

(You will notice that I did not use a WCPS request to get the min, mean, max values - instead I just summarized these from the regular WCS results. I could not seem to write a WCPS query that would return the correct data in this structure without resorting to a bunch of iteration wrapping the WCPS query, sending lots of round trip requests to rasdaman. This method seemed much simpler / more transparent to me, but of course I am open to suggestions!)

To test, run the flask application as described in `README.md`, then review the documentation at `/hydrology.html`. Then try:

Standard endpoint:

```
/hydrology/point/62/-145
```

Endpoints with ?summarize=mmm argument, ?format=csv argument, and combined ?summarize=mmm&format=csv arguments:

```
/hydrology/point/62/-145?summarize=mmm
/hydrology/point/62/-145?format=csv
/hydrology/point/62/-145?summarize=mmm&format=csv

```
Bad arguments to make sure no data is returned:

```
/hydrology/point/62/-145?summarize=mean
/hydrology/point/62/-145?format=xlsx
/hydrology/point/62/-145?summarize=mean&format=csv
/hydrology/point/62/-145?summarize=mmm&format=xlsx
```

### DIFF
--- a/routes/hydrology.py
+++ b/routes/hydrology.py
@@ -1,4 +1,5 @@
 import asyncio
+import numpy as np
 from urllib.parse import quote
 from flask import (
     Blueprint,
@@ -142,10 +143,9 @@ def run_fetch_hydrology_point_data(lat, lon):
     return point_pkg
 
 
-### unfinished function, need to fix WCPS request string
 def run_fetch_hydrology_point_data_mmm(lat, lon):
-    """Fetch summarized hydrology data for a
-       given latitude and longitude.
+    """Fetch hydrology data for a
+       given latitude and longitude, and summarize over eras.
 
     Args:
         lat (float): latitude
@@ -155,16 +155,71 @@ def run_fetch_hydrology_point_data_mmm(lat, lon):
         JSON-like dict of data at provided latitude and
         longitude for all variables
     """
-    x, y = project_latlon(lat, lon, 3338)
+    # get standard point data package
+    point_pkg = run_fetch_hydrology_point_data(lat, lon)
 
-    wcps_str = generate_average_wcps_str(x, y, hydrology_coverage_id, "eras", (0, 14))
-    wcps_query_url = generate_wcs_query_url(wcps_str)
-    rasdaman_response = asyncio.run(fetch_data([wcps_query_url]))
-    return rasdaman_response
+    # compute min/mean/max across eras for each variable name
+    mmm_dict = dict()
 
-    # package point data with decoded coord values (names)
-    # point_pkg = dict()
-    # return point_pkg
+    var_names = []
+    for var_coord in dim_encodings["varnames"].keys():
+        var_names.append(dim_encodings["varnames"][var_coord])
+
+        for var_name in var_names:
+            values = []
+
+            for model_coord in dim_encodings["models"].keys():
+                model_name = dim_encodings["models"][model_coord]
+                for scenario_coord in dim_encodings["scenarios"].keys():
+                    scenario_name = dim_encodings["scenarios"][scenario_coord]
+                    for month_coord in dim_encodings["months"].keys():
+                        month_name = dim_encodings["months"][month_coord]
+                        for era_coord in dim_encodings["eras"].keys():
+                            era_name = dim_encodings["eras"][era_coord]
+
+                            values.append(
+                                float(
+                                    point_pkg[model_name][scenario_name][month_name][
+                                        era_name
+                                    ][var_name]
+                                )
+                            )
+
+            min_value, mean_value, max_value = (
+                min(values),
+                round(np.nanmean(values), 2),
+                max(values),
+            )
+
+            mmm_dict[var_name] = min_value, mean_value, max_value
+
+    # repackage point data with mmm values
+    point_pkg_mmm = dict()
+    for model_coord in dim_encodings["models"].keys():
+        model_name = dim_encodings["models"][model_coord]
+        point_pkg_mmm[model_name] = dict()
+        for scenario_coord in dim_encodings["scenarios"].keys():
+            scenario_name = dim_encodings["scenarios"][scenario_coord]
+            point_pkg_mmm[model_name][scenario_name] = dict()
+            for month_coord in dim_encodings["months"].keys():
+                month_name = dim_encodings["months"][month_coord]
+                point_pkg_mmm[model_name][scenario_name][month_name] = dict()
+                for var_coord in dim_encodings["varnames"].keys():
+                    var_name = dim_encodings["varnames"][var_coord]
+                    point_pkg_mmm[model_name][scenario_name][month_name][
+                        var_name
+                    ] = dict()
+                    point_pkg_mmm[model_name][scenario_name][month_name][var_name][
+                        "min"
+                    ] = mmm_dict[var_name][0]
+                    point_pkg_mmm[model_name][scenario_name][month_name][var_name][
+                        "mean"
+                    ] = mmm_dict[var_name][1]
+                    point_pkg_mmm[model_name][scenario_name][month_name][var_name][
+                        "max"
+                    ] = mmm_dict[var_name][2]
+
+    return point_pkg_mmm
 
 
 @routes.route("/hydrology/")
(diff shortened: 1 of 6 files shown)


templates/documentation/hydrology.html
```diff
@@ -0,0 +1,258 @@
...
+          "glacier_melt": &lt;monthly sum of daily glacier ice melt, mm&gt;,
+          "iwe": &lt;monthly maximum of daily ice water equivalent, mm&gt;,
+          "pcp": &lt;monthly sum of daily precipitation, mm&gt;,
+          "runoff": &lt;monthly sum of daily surface runoff, mm&gt;,
+          "sm1": &lt;monthly mean of daily soil moisture in layer 1, mm&gt;,
+          "sm2": &lt;monthly mean of daily soil moisture in layer 2, mm&gt;,
+          "sm3": &lt;monthly mean of daily soil moisture in layer 3, mm&gt;,
+          "snow_melt": &lt;monthly sum of daily snowmelt, mm&gt;,
+          "swe": &lt;monthly maximum of daily snow water equivalent, mm&gt;,
+          "tmax": &lt;monthly mean of daily maximum air temperature at 2m, degrees C&gt;,
```
charparr, line 108:
```suggestion
          "tmax": &lt;monthly mean of daily maximum air temperature at 2m, &deg;C&gt;,
```

charparr: I think this will yield a "degrees" symbol

routes/hydrology.py
```diff
@@ -0,0 +1,307 @@
...
+                                )
+                            )
+
+            min_value, mean_value, max_value = (
+                min(values),
+                round(np.nanmean(values), 2),
+                max(values),
+            )
+
+            mmm_dict[var_name] = min_value, mean_value, max_value
```
charparr, line 194: This block might need to be indented another level? I'm currently fetching duplicate values across scenarios for a `mmm` query like this:

http://localhost:5000/hydrology/point/61.20663/-148.95676?summarize=mmm

image

I guess this could be happening in the packaging component too, but that part looked straightforward on my first reading.

charparr (changes requested):
Awesome endpoint! I've left a few minor suggestions, see my comment about duplicated values for mmm queries across scenarios for a requested change. Cool to see these data come to life via the API.

routes/hydrology.py
charparr: Oh on closer inspection these values are duped between models as well:
image

cstephen (changes requested):
This endpoint and documentation are looking great overall!

I noticed the same repeated mmm-summary values issue that @charparr already mentioned, but besides that, the only (minor) issue I found is that a bad `format` parameter returns an unformatted (non-template version) "Internal Server Error" page:

http://localhost:5000/hydrology/point/61.5/-147?format=badvalue

Compare this with the HTTP 400 "Bad Request" error page that is returned when a bad value is passed into the `summarize` parameter:

http://localhost:5000/hydrology/point/61.5/-147?summarize=badvalue

Expected behavior is that the bad value passed into the `format` parameter returns the same HTTP 400 "Bad Request" error page.

charparr:
There actually four citations / links I think we'll want here:

The original source dataset: this is the link that is currently in the doc page!

The paper about the original source dataset: which is this one https://doi.org/10.1016/j.cliser.2022.100312

The paper Jeremy is writing about this data as we are serving it (decadal monthly summaries): we can see if he has a title and add an  _(in prep) or (in review)_ or whatever.

The link to the dataset on on our own GeoNetwork Data Portal once it is live.

templates/documentation/hydrology.html
Joshdpaul: addressed with https://github.com/ua-snap/data-api/pull/358/commits/d22f5bf6bbd6cdd58d341a31c42b2f75f3ab724f

templates/documentation/hydrology.html
```diff
@@ -0,0 +1,258 @@
...
+      </td>
+    </tr>
+  </tfoot>
+</table>
+
+<h3>Output</h3>
+
+<h4>Decadal hydrology variables</h4>
+
+<p>Results from decadal queries will look like this:</p>
```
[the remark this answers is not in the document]
Joshdpaul: addressed with https://github.com/ua-snap/data-api/pull/358/commits/d22f5bf6bbd6cdd58d341a31c42b2f75f3ab724f

routes/hydrology.py
Joshdpaul: I found the problem and was able to incorporate the mmm data grab inside of the repackaging instead.... fixed the problem with https://github.com/ua-snap/data-api/pull/358/commits/82e4b3ba34b49aeb88bc009af7332e32065e82a2 by _removing_ code for a change :) 
image

routes/hydrology.py
```diff
@@ -0,0 +1,307 @@
+import asyncio
+import numpy as np
+from urllib.parse import quote
```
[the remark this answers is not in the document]
Joshdpaul: fixed with https://github.com/ua-snap/data-api/pull/358/commits/84b8a6b6637d6cf36acff62c6ef0010c70c24d76

routes/hydrology.py
```diff
@@ -0,0 +1,307 @@
...
+)
+
+# local imports
+from fetch_data import fetch_wcs_point_data, fetch_data
+from validate_request import (
+    validate_latlon,
+    project_latlon,
+)
+from generate_requests import generate_average_wcps_str
+from generate_urls import generate_wcs_query_url
```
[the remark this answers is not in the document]
Joshdpaul: fixed with https://github.com/ua-snap/data-api/pull/358/commits/84b8a6b6637d6cf36acff62c6ef0010c70c24d76

Joshdpaul:
> This endpoint and documentation are looking great overall!
> 
> I noticed the same repeated mmm-summary values issue that @charparr already mentioned, but besides that, the only (minor) issue I found is that a bad `format` parameter returns an unformatted (non-template version) "Internal Server Error" page:
> 
> http://localhost:5000/hydrology/point/61.5/-147?format=badvalue
> 
> Compare this with the HTTP 400 "Bad Request" error page that is returned when a bad value is passed into the `summarize` parameter:
> 
> http://localhost:5000/hydrology/point/61.5/-147?summarize=badvalue
> 
> Expected behavior is that the bad value passed into the `format` parameter returns the same HTTP 400 "Bad Request" error page.

Bad format args behavior fixed with https://github.com/ua-snap/data-api/pull/358/commits/c55666f2a22745ea32ca9a1987c91d642a9901e7

Joshdpaul:
As part of the revisions in https://github.com/ua-snap/data-api/pull/358/commits/82e4b3ba34b49aeb88bc009af7332e32065e82a2, I also incorporated a check for NaN in the min/mean/max values. This was because my JSONView add-on was now telling me that `{"max":NaN}` was not valid JSON. So I just changed it to the string "nan", but I realize this creates mixed data types in the mmm values. Is this OK? Do you have any suggestions that don't end up mixing floats and strings like this?

image

Joshdpaul:
> There actually four citations / links I think we'll want here:
> 
> The original source dataset: this is the link that is currently in the doc page!
> 
> The paper about the original source dataset: which is this one https://doi.org/10.1016/j.cliser.2022.100312
> 
> The paper Jeremy is writing about this data as we are serving it (decadal monthly summaries): we can see if he has a title and add an _(in prep) or (in review)_ or whatever.
> 
> The link to the dataset on on our own GeoNetwork Data Portal once it is live.

The first two have been incorporated into the documentation, and the last two have been punted into issue #363 since we cannot yet link to them

charparr (approved):
Looking good! I tested a bunch of `mmm` queries and am seeing the revised, non-duplicate results. The CSVs look really good for this endpoint, everything is explained very explicitly making the CSVs easy to parse independent of any external information source.

cstephen (approved):
I've confirmed that the http://localhost:5000/hydrology/point/61.5/-147?format=badvalue URL now returns the expected template-based error page. Approved!

PR #359, opened 2023-10-03
Changes the metadata for all CSVs to point at the URL to get the data from the API.

### LINKED ISSUE #355: Modify CSV to remove top line of metadata regarding Northern Climate Reports and replace with earthmaps.io endpoint
Instead of stating:
# View a report for this location at https://northernclimatereports.org/report/65/-155

We would want:
# View this data at https://earthmaps.io/<endpoint>/65/-155

BobTorgerson:
This PR changes the URL that is supplied in all of the CSV metadata that used to point at the NCR. It instead points at the data endpoint from which the CSV was generated.

Fixes #355

### DIFF
--- a/csv_functions.py
+++ b/csv_functions.py
@@ -1,4 +1,4 @@
-from flask import request, Response
+from flask import request, Response, request
 import csv
 from postprocessing import *
 from fetch_data import *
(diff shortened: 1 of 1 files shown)


cstephen (approved):
This is an excellent change now that the API is not so NCR-specific, and everything works as expected. Approved!

PR #370, opened 2023-10-05
Updates temperature summary with annual min, mean, and max

BobTorgerson:
This PR fixes the algorithm for generating the annual mean temperature summary where the min and max are taken from the yearly means rather than the maximum or minimum of the tasmax / tasmin respectively. 

These are giving values much closer to the expected annual values we saw in the previous iteration of the temperature summary for the ArcticEDS.

To test, you can run this branch and request:
http://localhost:5000/eds/temperature/64.8378/-147.716 

The above URL shows temperature data for Fairbanks.

### DIFF
--- a/routes/taspr.py
+++ b/routes/taspr.py
@@ -707,9 +707,7 @@ def create_temperature_eds_summary(temp_json):
     )
     hist_monthly_mmm = pd.DataFrame(columns=["Month", "tasmean", "tasmax", "tasmin"])
 
-    all_monthly_means = []
-    all_monthly_maxes = []
-    all_monthly_mins = []
+    all_monthly_means = list()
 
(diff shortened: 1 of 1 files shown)


cstephen (approved):
This PR looks good to me! I tested the following:
- I looked through the new code to make sure I could follow its logic.
- To test the data integrity of the Mean Annual Temperature table, I set the Mean Monthly Temperatures table below it to 5 Model Average + RCP 8.5, then added up the numbers and divided by 12 to manually calculate the mean annual temperature for each era, and the values matched the "Mean" column of the Mean Annual Temperature table (with very slight rounding differences).
- I looked at the Mean Annual Temperature table for several other locations to make sure the values looked sane.

PR #372, opened 2023-10-09
Adds EDS endpoint for Hydrology data

BobTorgerson:
This PR adds the ArcticEDS endpoint for the hydrology section of the report. This code generates the min-mean-max for each of the months in a given era such as historical being 1950-2009 and then takes the the decadal annual means to find the min-mean-max from those values. 

The summary contains those values in each of the models / scenarios as "Annual" which the ArcticEDS populates its tables from. In addition, it contains the CSV preview, similar to other EDS endpoints in the API currently.

Be sure to set your environment to Apollo:
`export RAS_API_BASE_URL=https://apollo.snap.uaf.edu/rasdaman/`

### DIFF
--- a/routes/hydrology.py
+++ b/routes/hydrology.py
@@ -5,6 +5,7 @@ from flask import (
     render_template,
     request,
     current_app as app,
+    jsonify
 )
 
 # local imports
@@ -220,7 +221,7 @@ def hydro_about():
 
 
 @routes.route("/hydrology/point/<lat>/<lon>")
-def run_get_hydrology_point_data(lat, lon):
+def run_get_hydrology_point_data(lat, lon, summarize=None, preview=None):
     validation = validate_latlon(lat, lon)
     if validation == 400:
         return render_template("400/bad_request.html"), 400
@@ -233,7 +234,7 @@ def run_get_hydrology_point_data(lat, lon):
         )
 
     # validate request arguments if they exist; set summarize argument accordingly
-    if len(request.args) == 0:
+    if len(request.args) == 0 and (summarize is None and preview is None):
         try:
             point_pkg = run_fetch_hydrology_point_data(lat, lon)
             return postprocess(point_pkg, "hydrology")
@@ -264,8 +265,8 @@ def run_get_hydrology_point_data(lat, lon):
         else:
             return render_template("400/bad_request.html"), 400
 
-    elif "summarize" in request.args:
-        if request.args.get("summarize") == "mmm":
+    elif "summarize" in request.args or summarize:
+        if request.args.get("summarize") == "mmm" or summarize:
             try:
                 return run_fetch_hydrology_point_data_mmm(lat, lon)
             except Exception as exc:
@@ -275,8 +276,8 @@ def run_get_hydrology_point_data(lat, lon):
         else:
             return render_template("400/bad_request.html"), 400
 
-    elif "format" in request.args:
-        if request.args.get("format") == "csv":
+    elif "format" in request.args or preview:
+        if request.args.get("format") == "csv" or preview:
             try:
                 point_pkg = run_fetch_hydrology_point_data(lat, lon)
                 place_id = request.args.get("community")
@@ -285,6 +286,7 @@ def run_get_hydrology_point_data(lat, lon):
                         point_pkg, "hydrology", place_id=place_id, lat=lat, lon=lon
                     )
                 else:
+                    print("Getting here")
                     return create_csv(
                         point_pkg, "hydrology", place_id=None, lat=lat, lon=lon
                     )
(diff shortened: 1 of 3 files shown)


routes/hydrology.py
```diff
@@ -264,19 +377,19 @@ def run_get_hydrology_point_data(lat, lon):
...
-                    return render_template("404/no_data.html"), 404
-                return render_template("500/server_error.html"), 500
+    elif "summarize" in request.args or summarize:
+        if request.args.get("summarize") == "mmm" or summarize:
+            # try:
+            return run_fetch_hydrology_point_data_mmm(lat, lon, summarize)
+            # except Exception as exc:
+            #     if hasattr(exc, "status") and exc.status == 404:
+            #         return render_template("404/no_data.html"), 404
+            #     return render_template("500/server_error.html"), 500
```
charparr, line 387: Uncomment this guy I think :)

charparr (changes requested):
This looks great, I tried a bunch of routes and verified that the data look as expected, and that for `summarize=mmm` variations there are no duplicate data values between different climate futures. The CSV responses look good as well. Here are some routes that I tried in testing:

Hydro previews and data for EDS
http://127.0.0.1:5000/eds/hydrology/65/-146

Hydro point data
http://127.0.0.1:5000/hydrology/point/65/-146

Hydro point data summarized, but no EDS-style preview
http://127.0.0.1:5000/hydrology/point/65/-146?summarize=mmm

Above, but in CSV
http://127.0.0.1:5000/hydrology/point/65/-146?summarize=mmm&format=csv

No summary CSV
http://127.0.0.1:5000/hydrology/point/65/-146?format=csv

I also tried the following routes and hit the appropriate status codes that I'd expect:
http://127.0.0.1:5000/hydrology/point/65/-146?format=foo
http://127.0.0.1:5000/hydrology/point/65/-146?summarize=foobar

See my one note to un-comment a try-except block, because I hit one route that should have probably been a 404 because ocean:
http://127.0.0.1:5000/hydrology/point/51.5/-146

routes/hydrology.py
BobTorgerson: Thanks for catching this Charlie! I've got the try-catch block uncommented now.

charparr:
Ah you know I see now after pulling that change that this is actually because `HadGEM2-ES` is 🙅  for `tmin`, `tmax`, and `pcp`. So this route http://127.0.0.1:5000/hydrology/point/51.5/-146 (middle of the Gulf of Alaska) gives a mix of `"nan"`s and `null`s (see screenshot below). This was noted previously, do we have an easy way to handle this with `postprocess`?

image

routes/hydrology.py
```diff
@@ -171,43 +179,148 @@ def run_fetch_hydrology_point_data_mmm(lat, lon):
...
+                                )
                             )
+
+                        # This will be the min, mean, and max taken from the mean monthly values
+                        # to find the min-mean-max of the annual values for a given time period such
+                        # as 1950-2009 for historical.
+                        min_value, mean_value, max_value = (
+                            min(values),
+                            round(np.nanmean(values), 2),
+                            max(values),
```
cstephen, line 293: I noticed that if you put a `print(len(values))` before this chunk of code, it shows that the `values` list has 12 values. I think this should be anywhere from 36-72 values (12 months * 3 decades for projected, or 12 months * 6 decades for historical), right?

charparr:
We may be able to just workaround it, with the below change that ocean route now 404s:

image

cstephen (changes requested):
This is looking great! I noticed a couple oddities though:

- [ ] I tried printing out the length of `values` right before the min/mean/max is taken in a couple places in the code, and I found one spot where the number of values seems to be less than what it should be (for the Arctic EDS annual min/mean/max). See the PR code comments for more details.
- [ ] I noticed when pointing the Arctic EDS webapp at this branch that the historical max value for Runoff changes when switching between models. This is the case for Fairbanks and several other locations I tested. Normally I would expect the historical values not to change at all, so I'm not sure what's going on here.

charparr:
Regarding this one:
> 
>     * [ ]  I noticed when pointing the Arctic EDS webapp at this branch that the historical max value for Runoff changes when switching between models. This is the case for Fairbanks and several other locations I tested. Normally I would expect the historical values not to change at all, so I'm not sure what's going on here.

The historical data here are actually historical runs of climate models, vs. reanalysis, so we would expect to see changes in the historical data when flipping between models and scenarios!

routes/hydrology.py
BobTorgerson: No, that should have only 12 values as those are the 12 mean values for each month for the particular "era" which is something like: historical, early_century, mid_century, and late century by that point in the script. That is the case due to the code in 184 - 224 that generates the monthly based mmm for each era.

BobTorgerson: That might be better clarified with some additional comments.

cstephen (approved):

charparr (approved):

PR #376, opened 2023-10-10
Fixes NCR permafrost endpoint by adding new conditional for grabbing required NCR data from Rasdaman.

### LINKED ISSUE #373: Permafrost endpoint needs to be retrofit to work with NCR
NCR was working with CRREL GIPL flavor and probably #354 introduced breaking changes to the structure of the returned data for NCR (but worked with EDS!).

What needs to happen is to

1. rewind in source history to recover how that endpoint was structured previously,
2. re-add that in some fashion (can be a non-beautiful route / not best documentation, we need it working more than we need it perfect) and re-attach that to the `ncr/` flavor of request (the route we need to work is i.e. `/ncr/permafrost/point/64.8378/-147.716`.)

BobTorgerson:
This PR fixes the NCR permafrost endpoint by requesting a different data volume from Rasdaman than the ArcticEDS does. This will return 3 models and 2 scenarios for the /ncr permafrost endpoint now, while keeping the ArcticEDS endpoint the same as it was previously.

Closes #373

### DIFF
--- a/routes/permafrost.py
+++ b/routes/permafrost.py
@@ -1,5 +1,6 @@
 import asyncio
 import pandas as pd
+import numpy as np
 from urllib.parse import quote
 from flask import Blueprint, render_template, request, jsonify, Response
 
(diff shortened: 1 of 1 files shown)


charparr (approved):
Looks good, I'm testing and getting non-duplicate, NCR-esque results as expected from routes like http://127.0.0.1:5000/ncr/permafrost/point/63.0628/-146.1627

It's good that this is under a "special" NCR route because we should remember that it is mmm`ing the data differently than normal, i.e. what is done by http://127.0.0.1:5000/permafrost/point/gipl/63.0628/-146.1627/2040/2069?summarize=mmm

PR #381, opened 2023-10-13
Adds seasonal totals for the NCR of the hydrology data

BobTorgerson:
This PR adds the seasonal totals for the hydrology data to display in the hydrology section of the NCR. 

You get these additional seasonal totals when requesting /eds/hydrology currently, which in a future version of the API we could change if desired. 

In the returned JSON, for each variable and scenario, you should see a seasonal total for Spring, Summer, Fall, and Winter.

### DIFF
--- a/routes/hydrology.py
+++ b/routes/hydrology.py
@@ -169,9 +169,6 @@ def run_fetch_hydrology_point_data_mmm(lat, lon, summarize=None):
                 point_pkg_mmm[model_name][scenario_name][month_name] = dict()
                 for var_coord in dim_encodings["varnames"].keys():
                     var_name = dim_encodings["varnames"][var_coord]
-                    point_pkg_mmm[model_name][scenario_name][month_name][
-                        var_name
-                    ] = dict()
 
                     # If summarizing for ArcticEDS, we want to get the min-mean-max for each
                     # month for the given era i.e. 1950-2009 for historical period.
@@ -181,6 +178,9 @@ def run_fetch_hydrology_point_data_mmm(lat, lon, summarize=None):
                             or var_name == "runoff"
                             or var_name == "sm1"
                         ):
+                            point_pkg_mmm[model_name][scenario_name][month_name][
+                                var_name
+                            ] = dict()
                             for era_title in dim_encodings["eds_eras"].keys():
                                 values = list()
                                 point_pkg_mmm[model_name][scenario_name][month_name][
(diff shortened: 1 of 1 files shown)


charparr (changes requested):
I'm just going to narrate how I'm reading this:

For each model, scenario, season, and variable, and era

Compute the average value across all occurents of that month in the set above (basically across all years in the era).

These average monthly values are then summed together on a seasonal basis to compute the average seasonal total for a specific model/scenario/variable combination.

This makes sense for what we want for runoff and evap, and matches what I'm seeing in the JSON repsonse for routes like 

http://127.0.0.1:5000/eds/hydrology/point/66/-147

and

http://127.0.0.1:5000/eds/all/66/-147.

However, summing each variable may not be the right path, and I should have made this clear earlier!

For variables that are already monthly sums, summing the monthly averages into a seasonal total makes sense. These variables are the "water flux" sort:
 - evap
 - runoff
 - snow melt
 - glacier melt
 - precip (pcp)

For the variables that are monthly summarized by a mean value, we'd want seasonal means there as well. Otherwise these values will be too big! Like a summer tmax of 55 C! That'd be 131 F for Interior Alaska. 
So, for the following variables we should have seasonal averages, not totals:
 - tmin
 - tmax
 - SM1
 - SM2
 - SM3

Finally, we have the variables that are monthly summarized by finding the maximum value. We'd want probably want to seasonally max these as well, i.e. "Maximum Spring Snow-Water-Equivalent" or similar. That is these variables.
 - SWE
 - IWE

I realize that we are only doing evap and runoff in the clients at the moment, if/when we implement other variables seasonally, we'll now have the correct functions / data results in place already.

We may just be able to handle this a dict that holds the seasonally summarizing function:
```python
def seasonal_sum(values):
    return round(np.sum(values), 0)


def seasonal_avg(values):
    return round(np.mean(values), 0)


def seasonal_max(values):
    return round(np.max(values), 0)


seasonal_summary_func = {
  "evap": seasonal_sum,
  "glacier_melt": seasonal_sum,
  "iwe": seasonal_max,
  "pcp": seasonal_sum,
  "runoff": seasonal_sum,
  "sm1": seasonal_avg,
  "sm2": seasonal_avg,
  "sm3": seasonal_avg,
  "snow_melt": seasonal_sum,
  "swe": seasonal_max,
  "tmax": seasonal_avg,
  "tmin": seasonal_avg,
}
```
Then you could presumably construct the total value like this:
```python
total_value = seasonal_summary_func([var_name], values))
```
That might not be the correct syntax, but I think it is close. This might also be a total Friday-brained thing, so if it doesn't make sense please let me know!!

cstephen (approved):
Looks good to me. I ran a couple manual calculations to make sure the means from the three months of a season (JJA and DJF) match their seasonal totals (Summer and Winter). Sounds like these values may change to seasonal means or maxes in some cases after considering @charparr's feedback, but the month/season alignment will still be good regardless 👍

charparr (approved):

charparr:
Hey I looked at the most recent commits and this is good! I tried a few routes and can confirm that the only variables present are evap and runoff. This matches what we expect in the clients at the moment. The `preview` piece of the JSON looks good still (all vars). Craig thanks for checking the seasonal math! Merging.

PR #384, opened 2023-10-16
fix no data handling

### LINKED ISSUE #379: 500 error in Indicators route
http://development.earthmaps.io//indicators/base/point/52.9365/173.24

What's going on here?  Could this be happening at other places, too?  Is this an unhandled no-data?

charparr:
This PR resolves an unhandled 500 in the `indicators` route when a lat-lon was supplied that is inside the BBOX, but is "off the map" in the sense that there is no data available for that specific point. This is often triggered when a marine lat-lon is used because most of our data is terrestrial.

I added a second `except` block that seems to resolve the issue, cleaned up some un-used imports, an updated a docstring while I was in there.

Try the following routes and outcomes:

http://127.0.0.1:5000/indicators/base/point/foo/-143.24    should yield Bad Request
http://127.0.0.1:5000/indicators/base/point/65/-143.24     should yield valid data JSON
http://127.0.0.1:5000/indicators/base/point/45/-143.24.    should yield Invalid Coordinates
http://127.0.0.1:5000/indicators/base/point/52.9365/173.24    should yield 404 No Data

I tried the coordinates from the above 404 route in a `permafrost/point` and a `hydrology/point` query and those both 404'd as expected, so I think this issue was confined to this particular route.

Closes #379

### DIFF
--- a/routes/indicators.py
+++ b/routes/indicators.py
@@ -6,8 +6,7 @@ These endpoint(s) query a coverage containing summarized versions of the indicat
 import asyncio
 import numpy as np
 from math import floor
-from flask import Blueprint, render_template, request, Response
-from shapely.geometry import Point
+from flask import Blueprint, render_template
 
 # local imports
 from generate_urls import generate_wcs_query_url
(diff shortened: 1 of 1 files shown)


cstephen (approved):
The URLs in the PR description behave as expected. Looks good 👍

PR #385, opened 2023-10-16
Changes the precipitation frequency CSV to have exceedance probability.

### LINKED ISSUE #382: Switch CSV download for Precip Frequency to exceedance probability
https://www.usgs.gov/centers/new-jersey-water-science-center/floods-recurrence-intervals-and-100-year-floods

For the downloaded CSV:
- Change the documentation row `# return_interval is used for the Annual Exceedance Probability (1 / return_interval)` to read `# exceedance_probability = annual exceedance probability`
- Rename column header `return_interval` to `exceedance_probability`
- Change the results in that column to be the exceedance probability (basically 100/x).

BobTorgerson:
This PR changes the precipitation frequency CSV to have exceedance probability rather than return interval. This simply checks if a CSV is being requested, and modifies the value that is being stored in that column.

Closes #382

### DIFF
--- a/csv_functions.py
+++ b/csv_functions.py
@@ -693,11 +693,11 @@ def taspr_csv(data, endpoint):
             filename_data_name = "Precipitation"
 
     elif endpoint == "proj_precip":
-        coords = ["return_interval", "duration", "model", "era"]
+        coords = ["exceedance_probability", "duration", "model", "era"]
         values = ["pf", "pf_lower", "pf_upper"]
         fieldnames = coords + values
         csv_dicts = build_csv_dicts(data, fieldnames, values=values)
-        metadata = "# return_interval is used for the Annual Exceedance Probability (1 / return_interval)\n"
+        metadata = "# exceedance_probability is the annual exceedance probability\n"
         metadata += "# duration is the amount of time for the predicted amount of precipitation\n"
         metadata += "# model is the model the data is derived from\n"
         metadata += (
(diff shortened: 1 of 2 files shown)


cstephen (approved):
This looks good to me. I've verified that the numbers that show up in the new exceedance_probability column are 100 / x, where x is the frequency value from the corresponding JSON output.

csv_functions.py
```diff
@@ -693,11 +693,11 @@ def taspr_csv(data, endpoint):
             filename_data_name = "Precipitation"
 
     elif endpoint == "proj_precip":
-        coords = ["return_interval", "duration", "model", "era"]
+        coords = ["exceedance_probability", "duration", "model", "era"]
         values = ["pf", "pf_lower", "pf_upper"]
         fieldnames = coords + values
         csv_dicts = build_csv_dicts(data, fieldnames, values=values)
-        metadata = "# return_interval is used for the Annual Exceedance Probability (1 / return_interval)\n"
+        metadata = "# exceedance_probability is the annual exceedance probability\n"
```
charparr, line 700: I think we should add the unit of "percent" to the exceedance probability value for clarity. 
```suggestion
        metadata = "# exceedance_probability is the annual exceedance probability in percent\n"
```

charparr (changes requested):
See my one suggestion for adding a unit of "percent" to the csv metadata. Otherwise I like this change, it helps clarify what these data are and provides better alignment with the app and technical report as well.

charparr (approved):

PR #391, opened 2023-10-18
Removes commented out elif block.

### LINKED ISSUE #387: Commented out `elif` block in main branch?
What happened here? Can we test + remove?

https://github.com/ua-snap/data-api/blob/b17dedc2783921ddb7219f816eb056546e157b36/routes/taspr.py#L1508-L1509

BobTorgerson:
Closes #387


charparr (approved):
Looks good and as expected this had no impact on the route under which this block was nested, e.g., http://127.0.0.1:5000/temperature/july/65/-145/2030/2040

PR #394, opened 2023-10-18
Add CSV export to indicators endpoints

### LINKED ISSUE #378: Add CSV format export to Indicators
Doesn't look like `?format=csv` is implemented there yet.

cstephen:
Closes #378.

This PR adds CSV export support for point and area indicators endpoints.

This required more changes than I anticipated because the indicators JSON output had some unnecessary era/model/scenario combinations (e.g., `historical` era with scenario `rcp45`) that always contained nodata and that, after being processed with our `nullify_and_prune` function, ended up as partial branches with null values (which is what `nullify_and_prune` is designed to do). This doesn't cause any problems on the webapp side, but because the functions we use to construct CSV rows expect a consistent level of key/value nesting, the functions didn't know what to do with the partial JSON branches.

To solve this, I've implemented some pruning of the JSON data in the `indicators.py` route before the data reaches the CSV functions, following a few simple rules. I.e., do not include a branch in the indicators JSON data if:

- The era is `historical` and the scenario is not `historical`
- The era is not `historical` and the scenario is `historical`
- The era is `historical` and the model is not `Daymet`
- The era is not `historical` and the model is `Daymet`

This eliminates the unnecessary branches in both the JSON output and the CSV rows, and fixes the incompatibility with our CSV export functions.

To test:

- Try downloading an indicators CSV for a point endpoint, (e..g, http://localhost:5000/indicators/base/point/64.5288/-146.899?format=csv)
- Try downloading an indicators CSV for an area endpoint (e.g., http://localhost:5000/indicators/base/area/19090108?format=csv)
- (For @BobTorgerson) Point the NCR webapp at the API and make sure the indicators tables continue to populate as expected for both points and areas since the JSON data has been cleaned up a bit.

### DIFF
--- a/routes/indicators.py
+++ b/routes/indicators.py
@@ -69,7 +69,7 @@ def package_base_indicators_data(point_data_list):
                 # Skip impossible combinations of era and model.
                 if era == "historical" and model != "Daymet":
                     continue
-                if era != "historical" and model == "Daymet":
+                elif era != "historical" and model == "Daymet":
                     continue
                 di[indicator][era][model] = dict()
                 for si, stat_li in enumerate(scenario_li):
@@ -77,7 +77,7 @@ def package_base_indicators_data(point_data_list):
                     # Skip impossible combinations of era and scenario.
                     if era == "historical" and scenario != "historical":
                         continue
-                    if era != "historical" and scenario == "historical":
+                    elif era != "historical" and scenario == "historical":
                         continue
                     di[indicator][era][model][scenario] = dict()
                     for ti, value in enumerate(stat_li):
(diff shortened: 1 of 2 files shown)


charparr (approved):
This looks great, and the JSON is a lot cleaner now for non-CSV requests. Thanks Craig! It was easy to follow what was happening here. I tested a few different CSV and non-CSV routes, for areas and points, and everything looks good. These CSVs are well documented with descriptions of what each indicator is.

routes/indicators.py
```diff
@@ -64,9 +66,19 @@ def package_base_indicators_data(point_data_list):
             di[indicator][era] = dict()
             for mi, scenario_li in enumerate(model_li):
                 model = base_dim_encodings["model"][mi]
+                # Skip impossible combinations of era and model.
+                if era == "historical" and model != "Daymet":
+                    continue
+                elif era != "historical" and model == "Daymet":
+                    continue
```
BobTorgerson, line 73: Good thinking! Removes unnecessary cruft from the resulting JSON and fewer loops made through the code

BobTorgerson (approved):
This all looks good to me and thanks for implementing the CSV functionality for these indicators!

PR #402, opened 2023-10-19
Remove extended-charset mm glyph

brucecrevensten:
Testing:

 - Trigger a CSV download for an Indicators endpoint, such as: `localhost:5000//indicators/base/point/64.8378/-147.716?format=csv&community=AK124`

Check that there's no goofy special characters in the metadata header at the top.


charparr:
Looks good to me! I checked a CSV, didn't see any goofed characters. It is probably good to be plain in the CSV for when we/others interact with them computationally as well, like `df = pd.read_csv`...

charparr (approved):

PR #419, opened 2024-02-20
Adds the CMIP6 indicators as a point query in data API

BobTorgerson:
This PR adds the CMIP6 indicator data available in Rasdaman as a new endpoint in the indicators endpoint.

To access the new indicators endpoint, you can request a point location by specifying a latitude and longitude where longitude is specified as 0-360° like: [http://localhost:5000/indicators/cmip6/point/61.5/247](http://localhost:5000/indicators/cmip6/point/61.5/247)

This update contains the ability to generate a CSV document of the returned data as well such as: [http://localhost:5000/indicators/cmip6/point/61.5/247?format=csv](http://localhost:5000/indicators/cmip6/point/61.5/247?format=csv)

Finally, there is some additional documentation that highlights the new CMIP6 indicators that have been added to the top of the indicators documentation at: [http://localhost:5000/indicators/](http://localhost:5000/indicators/). 

Please let me know if you think there is anything missing or that I should change!

### DIFF
--- a/routes/indicators.py
+++ b/routes/indicators.py
@@ -12,7 +12,12 @@ from flask import Blueprint, render_template, request
 from generate_urls import generate_wcs_query_url
 from generate_requests import generate_wcs_getcov_str
 from fetch_data import *
-from validate_request import validate_latlon, project_latlon, validate_var_id
+from validate_request import (
+    validate_latlon,
+    validate_cmip6_indicators_latlon,
+    project_latlon,
+    validate_var_id,
+)
 from postprocessing import nullify_and_prune, postprocess
 from csv_functions import create_csv
 from . import routes
@@ -21,9 +26,24 @@ from config import WEST_BBOX, EAST_BBOX
 indicators_api = Blueprint("indicators_api", __name__)
 # Rasdaman targets
 indicators_coverage_id = "ncar12km_indicators_era_summaries"
+cmip6_indicators_coverage_id = "cmip6_indicators"
 
 # dim encodings for the NCAR 12km BCSD indicators coverage
 base_dim_encodings = asyncio.run(get_dim_encodings(indicators_coverage_id))
+cmip6_dim_encodings = asyncio.run(get_dim_encodings(cmip6_indicators_coverage_id))
+
+
+async def fetch_cmip6_indicators_point_data(lat, lon):
+    """
+    TODO: Write the documentation for this function
+    """
+    wcs_str = generate_wcs_getcov_str(
+        lon, lat, cov_id=cmip6_indicators_coverage_id, projection="EPSG:4326"
+    )
+    url = generate_wcs_query_url(wcs_str)
+    point_data_list = await fetch_data([url])
+
+    return point_data_list
 
 
 # Base indicators endpoint:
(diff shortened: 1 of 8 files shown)


Joshdpaul (approved):
I was able to successfully access the newly ingested CMIP6 indicators data via the `.../indicators` endpoint. Both the CSV download and standard JSON-in-browser formats worked for me. The basic functionality is there. Approved!

**Comments:**
I found the actual values produced to be pretty wacky. @BobTorgerson confirmed this is a known issue, maybe caused by the Rasdaman ingest process. The values are far outside of the QC'd value range and so the issue is likely downstream from the prefect flow that computes the indicators.

One improvement I'd suggest is that we accept the coordinates input in standard format (eg, longitude -180 to 180) to match the rest of the endpoints. Even though the indicators data is in 0-360 format, we can probably convert this for the user since its so uncommon to use 0-360 longitude.

I tried some bad coordinates, and mostly got the expected "bad request" error messages. However, trying something like `http://127.0.0.1:5001/indicators/cmip6/point/-50/234` gives me the following bbox error. The text in this error does not really apply to the 0-360 range currently used for this endpoint, and also would not apply if we are serving the CMIP6 data in a Pan-Arctic domain. This should be revisited depending on the domain / coordinate format we decide on.

image

templates/documentation/indicators.html
```diff
@@ -14,7 +14,67 @@ <h2>Climate Indicators</h2>
...
+<table class="endpoints">
+  <thead>
+    <tr>
+      <th class="endpoint-label">Endpoint</th>
+      <th class="endpoint-url">Example URL</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>Annual mean values for all climate indicator variables at a given latitude and longitude.</td>
```
kyleredilla, line 32:
```suggestion
      <td>Yearly climate indicator values at a given latitude and longitude.</td>
```
Just a bit more precise, as these are not annual means

routes/indicators.py
```diff
@@ -101,6 +186,51 @@ def about_indicators():
...
+        place_id = request.args.get("community")
+        return create_csv(results, "cmip6_indicators", place_id, lat, lon)
+
+    return results
+
+    # except ValueError:
+    #     return render_template("400/bad_request.html"), 400
+    # except Exception as exc:
+    #     if hasattr(exc, "status") and exc.status == 404:
+    #         return render_template("404/no_data.html"), 404
```
kyleredilla, line 231: Just double checking whether we want to keep or remove this block

templates/documentation/indicators.html
kyleredilla, line 1: The intro text at the top of this page refers to the NCAR indicators only, I think that text should be changed to something more general that applies to both sets of indicators. If you want to keep the time range info, then move that to the NCAR section, and add to the CMIP6 section that the range of valid data goes from 1850-2100.

We will also need to add references for the CMIP6 data, but we don't have that ready yet, because I think we will want to reference the "regridded" dataset, instead of going through and citing each of the modelling groups for this derived product. So I think we can leave that alone for now.

kyleredilla (changes requested):
This is looking great. The CMIP6 indicators endpoint appears to be working as expected. I did see some things that could use some more work.

The main thing is to see if the NCAR indicators endpoint is still working for you, it appears to have stopped working for me, e.g. [http://localhost:5000/indicators/base/point/61.5/-147](http://localhost:5000/indicators/base/point/61.5/-147) gives the following:

image

I made some other suggestions as well, but they are minor.

templates/documentation/indicators.html
```diff
@@ -14,7 +14,67 @@ <h2>Climate Indicators</h2>
...
+{
+  "historical": {
+    EC-Earth3-Veg": {
+      "1850": {
+        "dw": 71,
+        "ftc": 23,
+        "rx1day": 9,
+        "su": 0
+      },
+      1851": {
```
brucecrevensten, line 62:
```suggestion
      "1851": {
```

templates/documentation/indicators.html
```diff
@@ -14,7 +14,67 @@ <h2>Climate Indicators</h2>
 
 <h3>Service endpoints</h3>
 
-<h4>All climate indicator variables</h4>
+<h4>CMIP6 climate indicator variables</h4>
+
+<p>
+  Query CMIP6 climate indicator variables for all models and scenarios.
+</p>
+
```
brucecrevensten, line 22: I suggest adding in here what the indicators themselves actually are / mean.  Like a simple list with the definitions of the indicators.

brucecrevensten: NM it's at the bottom!  But how about here, add a link (anchor to bottom) where it says just "See below for definitions of indicators included here."

templates/documentation/indicators.html
brucecrevensten: Yes: I think it would be most honest right now to explicitly state that the CMIP6 data will have peer-reviewed literature but it's not available yet, and they can email us with questions.

Joshdpaul:
The wacky values issue mentioned above seems to have been resolved after @BobTorgerson 's modification of the Rasdaman ingest of the `ncar12km_indicators_era_summaries`.

BobTorgerson:
I committed both of your suggested changes @kyleredilla! Thanks for reading through all of that! We may want to hold off on merging until we have the newest data with the longitude configured by default from -180 to 180 as I have a placeholder piece of code to do the conversion as the first step of the API endpoint.

PR #423, opened 2024-03-06
Adds map to API 422 returns for invalid latitude / longitude requests and shows the described bounding box on the map.

### LINKED ISSUE #420: Include a map with 422 bbox error
This error message could benefit from a simple map image showing the "east" and "west" bbox, to improve the user's understanding and of the spatial domain. (This is the `templates/422/invalid_latlon.html` error.)

image

BobTorgerson:
This PR adds a simple Leaflet map to the API's 422 HTTP code returns when an invalid latitude / longitude has been requested. The map has the bounding box that is described in the returned page to more quickly highlight where our data boundaries are for our point based queries.

An example of this error can be found at: http://localhost:5000/temperature/point/80.0628/-146.162

Note: The color of the bounding box may need to be adjusted to be more visible, but let me know if you think that is the case.

Closes #420

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -32,3 +32,8 @@ th.endpoint-url {
 table.data-source th {
 	white-space: nowrap;
 }
+
+#map {
+	height: 400px;
+	width: 400px;
+}
--- a/templates/422/invalid_latlon.html
+++ b/templates/422/invalid_latlon.html
@@ -30,9 +30,7 @@
     var map = L.map('map', {zoomControl: false, scrollWheelZoom: false}).setView([61.0, -151.505], 5);
     map.attributionControl.setPrefix('');
     
-    L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
-        maxZoom: 19,
-    }).addTo(map);
+    L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png').addTo(map);
 
     var southWest = L.latLng(51.3492, -187.5799),
         northEast = L.latLng(71.3694, -122.8098),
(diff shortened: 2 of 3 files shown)


Joshdpaul (approved):
Approved! This new feature works as expected.

I think the bounding box color is OK, I can tell what it is. But I'd be curious in seeing what something more colorful looks like (an orange or purple or something thats not so similar to the blue/white on the basemap).

kyleredilla (commented):
Huh, I am getting an "internal server error" for a couple of different points, inside both the "eastern" and "western" bounding boxes.

Neither of these work on the production API though, so maybe this is a separate issue I should ticket?

This one is inside the "western" bb:
[http://localhost:5000/temperature/point/65/-170](http://localhost:5000/temperature/point/65/-170)

And the "eastern" bb
[http://localhost:5000/temperature/point/65/175](http://localhost:5000/temperature/point/65/175)

It seems as though the invalid coordinates map + page correctly shows up whenever either or both of lat / lons are outside of the either bbox, so I would approve if you agree this looks like a separate issue I am seeing. 

Also agree with Josh on the bbox map color - it is definitely visible and useful, but a higher contrast color could be nice.

BobTorgerson:
@Joshdpaul @kyleredilla Thanks for your reviews! I changed the color to a dark orange that better highlights the boundaries and I think looks good. Have one more look at that.

As for the internal server error issues, that appears to be a separate issue that I'm going to make a Github issue for now. So we shouldn't hold this PR back because of that issue, as it was not introduced with any of this code.

kyleredilla (approved):
I do like that color bbox a lot more! That one seems like the "standard" to me, see it elsewhere.

Joshdpaul:
Yeah the orange looks a lot better! 👍🏻

PR #430, opened 2024-04-25
Degree days: replace all instances of "historical" with "modeled baseline"

charparr:
This PR is motivated by Arctic-EDS feedback that "historical" in the context of "historical modeled data" was causing the audience (engineers) to stub their toes. Generally this audience is less familiar with climate model data and they interpreted the "historical" label as "historical data must match observed station data."  

We're using the phrase "modeled baseline" instead.

To test this PR, make sure your API is pointing toward Zeus.

Then try some of the routes [listed here](https://github.com/ua-snap/data-api/pull/428) and verify that keys and labels all read "modeled baseline". This key is often nested under "daymet".

Also review the documentation template to verify that the references to "historical" data are also updated: http://localhost:5000/degree_days/

This PR also removes some commented-out code that was leftover from development that wasn't caught in the first iteration.

### DIFF
--- a/routes/degree_days.py
+++ b/routes/degree_days.py
@@ -23,6 +23,9 @@ degree_days_api = Blueprint("degree_days_api", __name__)
 
 # all degree day coverages share common dim_encodings, so only fetch one
 dd_dim_encodings = asyncio.run(get_dim_encodings("heating_degree_days_Fdays"))
+# update the encoding for "historical" to be "modeled baseline"
+# this is to make the data better align with engineer expectations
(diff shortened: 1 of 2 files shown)


cstephen (changes requested):
This PR looks good overall. I found a couple things to change though:

- [ ] One instance of a repeated word in the template documentation (see PR code comments)
- [ ] We decided to use `modeled_baseline` (with an underscore) instead of `modeled baseline` (with a space) for both the JSON and CSV output.

cstephen (approved):

PR #433, opened 2024-04-25
Remove erroneous scenario column from degree days mmm summary CSVs

### LINKED ISSUE #432: Extra/erroneous "scenario" column in degree days CSVs
CSV URLs like the following all contain the same erroneous `scenario` column, the values of which are always `ddmin`:

http://localhost:5000/degree_days/heating/66.8/-144.1?format=csv&summarize=mmm
http://localhost:5000/degree_days/heating/66.8/-144.1/1990/2040?format=csv&summarize=mmm
http://localhost:5000/degree_days/below_zero/66.8/-144.1?format=csv&summarize=mmm
http://localhost:5000/degree_days/below_zero/66.8/-144.1/1990/2040?format=csv&summarize=mmm
http://localhost:5000/degree_days/freezing_index/66.8/-144.1?format=csv&summarize=mmm
http://localhost:5000/degree_days/freezing_index/66.8/-144.1/1990/2040?format=csv&summarize=mmm
http://localhost:5000/degree_days/thawing_index/66.8/-144.1?format=csv&summarize=mmm
http://localhost:5000/degree_days/thawing_index/66.8/-144.1/1990/2040?format=csv&summarize=mmm

See the screenshot below for example:

image

It looks like this issue was introduced recently since the production API does not have this issue. Comparing the URLs above to their production counterparts, it looks like all we need to do here is just drop the `scenario` column entirely.

cstephen:
Closes #432.

This PR fixes the issue described in issue #432. To test, try each of the URLs in #432 and make sure the bad `scenarios` column has been removed from the CSVs. You may need to use a different browser or a private window to see the difference if you have visited the same URLs recently. Chrome caches CSV downloads very aggressively! Not sure about other browsers.

Also, spot check some non-mmm degree days CSV URLs to make sure this change didn't impact those. They look good to me, though.

### DIFF
--- a/csv_functions.py
+++ b/csv_functions.py
@@ -290,7 +290,7 @@ def degree_days_csv(data, endpoint):
         "air_thawing_index_Fdays",
         "air_freezing_index_Fdays",
     ]:
-        coords = ["model", "scenario"]
+        coords = ["model"]
         values = ["ddmin", "ddmean", "ddmax"]
     elif endpoint in [
         "heating_degree_days_Fdays_all",


charparr:
Good catch here, and this seems to fix it! I spot-checked a few summarized and unabridged CSVs and they looked good.

Issue #350, opened 2023-09-25
<var_ep> in taspr.py doesn't handle invalid requests


charparr:
Trying a route like `/foo/point/` unintentionally pings the taspr route below:

https://github.com/ua-snap/data-api/blob/cb282a5702c04ae00907c6ba272e332d97493fe6/routes/taspr.py#L1144-L1178

and instead of a `Bad Request` the result is a messy error when the packaging function fails.

charparr:
Closed by #445

PR #445, opened 2024-05-16
Modifies temperature, precipitation, and taspr routes to not use a wildcard for the endpoint.

BobTorgerson:
This PR expands the route definitions for temperature, precipitation, and taspr to no longer have a wildcard "catch all" start to the endpoint. Without this change, any non-existent endpoint that ends in point/<lat>/<lon> or area/<areaID> was resolving to these endpoints and causing an uncaught error to appear in Varnish (HTTP 503 error).

Example failure:

https://earthmaps.io/foo/bar/point/65/-156

If you run this code, and attempt that endpoint, it correctly gives a "bad request" error message:

http://localhost:5000/foo/bar/point/65/-156

Ensure that you can still get access to the temperature, precipitation, and taspr endpoints for both lat / lon and area summaries.

http://localhost:5000/temperature/point/65.0628/-146.1627
http://localhost:5000/precipitation/point/65.0628/-146.1627
http://localhost:5000/taspr/point/65.0628/-146.1627

http://localhost:5000/temperature/area/19010208
http://localhost:5000/precipitation/area/19010208
http://localhost:5000/taspr/area/19010208

### DIFF
--- a/routes/taspr.py
+++ b/routes/taspr.py
@@ -532,13 +532,13 @@ def package_mmm_point_data(point_data, cov_id, varname, start_year=None, end_yea
                             continue
                     point_pkg[dim_model][dim_scenario][str(year)] = dict()
                     if cov_id == "annual_precip_totals_mm":
-                        point_pkg[dim_model][dim_scenario][str(year)][
-                            "pr"
-                        ] = point_data[model][year_offset][scenario]
+                        point_pkg[dim_model][dim_scenario][str(year)]["pr"] = (
+                            point_data[model][year_offset][scenario]
+                        )
                     elif cov_id == "annual_mean_temp":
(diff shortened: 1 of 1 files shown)


kyleredilla (approved):
Yeah I see what's going on here, looks like these changes will fix the issue. Tested with supplied links in PR and cross-checked values with production API, things look good. Approved.

charparr (approved):
Hey this looks good! I'm testing the suggested endpoints and a few others and not hitting any issues - they work as expected for both cases where I anticipate a 503 and where I expect to get data. This PR will make it easier for us to test / develop new endpoints because we won't get misled by the mysterious guru meditation when in reality all we did was goof route the name. Like `temprature` which I type more than I'd like to admit. Thanks Bob!

charparr:
Closes #350

Issue #137, opened 2022-03-09
Removal of route synonyms


BobTorgerson:
We likely want to remove route synonyms from our API routes as a part of housecleaning. For example, we discovered that due to the permafrost.py routes containing: 
`@routes.route("/permafrost/protectedarea/")
@routes.route("/groundtemperature/protectedarea/")
@routes.route("/activelayer/protectedarea/")
@routes.route("/magtalt/protectedarea/")
@routes.route("/alt/protectedarea/")
@routes.route("/magt/protectedarea/")`

But only the /permafrost/protectedarea/ being defined, my code for re-caching Varnish was looking at the undefined endpoints and attempting to run them. This normally wouldn't have been anything but a 404 or possibly 500 error as those are not defined, but due to the fact that we have a variable starting endpoint using for /temperature, /precipitation, and /taspr in the taspr.py https://github.com/ua-snap/data-api/blob/f9130b2f43b19f5c9a8df9fb898ad3396b63158e/routes/taspr.py#L732
we are seeing those undefined routes attempted to be run for all of those additional routes above and hitting 503 errors when things go wrong (as they are not expected to use those routes).

Here is an example of that: http://localhost:5000/activelayer/huc/19060202

Bruce suggests we simplify as much as possible and try not to have too many "synonyms" for the same function where it makes sense.

charparr:
Close this ticket when we have verified that the following endpoints have been _de-synonym'd_ 
- [ ] Wildfire
- [ ] Climate Indicators
- [ ] Climate Protection from Spruce Beetles
- [ ] Degree Days
- [ ] Demographics
- [ ] Digital Elevation Models (DEMs)
- [ ] Flammability and Vegetation Type (ALFRESCO)
- [ ] Hydrology
- [ ] Landfast Sea Ice
- [ ] Permafrost
- [ ] Physical and Administrative Boundary Polygons
- [ ] Physiography
- [ ] Sea Ice Concentration
- [ ] Snowfall Equivalent
- [ ] Temperature and Precipitation
- [ ] Wet Days Per Year
- [ ] Wildfire

cstephen:
Xref: https://github.com/ua-snap/data-api/issues/618, sorta.

PR #462, opened 2024-10-02
Adds the WFS query for generating nearby fires to the /fire endpoint

### LINKED ISSUE #424: Add near-real-time fire, hotspot, smoke data endpoint
Add:

 * `/fire/point/[latlng]` - returns a list of all fire data within a radius (10mi? 50mi? ask) including AQI and other stuff that drives the wildfire map

Closes #288

### LINKED ISSUE #446: Add "active fires within ~70 miles" to return /fire/point endpoint using WFS
We would like to have an additional section in the /fire/point endpoint that returns any active fire points / polygons that intersect with the area of a 50 mile circle from a given point.

BobTorgerson:
This PR adds the WFS search query that finds all fires within ~70 miles of the selected point and returns the points and polygons inside of two additional JSON keys: fire_points and fire_polygons. To test, try points around the state of Alaska to see the returned data and ensure that the documentation for this returned data is clear. 

This also removes an unused endpoint /wildfire that has no purpose remaining available.

Closes #424 
Closes #446

### DIFF
--- a/generate_urls.py
+++ b/generate_urls.py
@@ -42,6 +42,10 @@ def generate_wfs_search_url(
     """
     distance = "0.7"
     if nearby_fires:
+        ''' 
+        GeoServer only supports degrees, so this is a query for ~70 mile radius.
+        https://gis.stackexchange.com/questions/132251/dwithin-wfs-filter-is-not-working
+       '''
         distance = "1.0"
     wfs_url = (
         GS_BASE_URL
--- a/generate_urls.py
+++ b/generate_urls.py
@@ -25,10 +25,27 @@ def generate_base_wfs_url(backend, workspace, lat, lon):
     return wfs_base
 
 
-def generate_wfs_search_url(workspace, lat, lon, hucs_pa_only=False):
+def generate_wfs_search_url(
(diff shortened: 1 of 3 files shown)


charparr (commented):
@BobTorgerson this looks good! I'm getting expected results for different locations around Alaska. Cool to have the proximity search baked in here.

My only question is that when I intentionally try a route where there will not be data like this:

http://127.0.0.1:5000/fire/point/53.0628/-146.1627

I get this response:
```
{
  "aqi_12": null,
  "aqi_24": null,
  "aqi_48": null,
  "aqi_6": null,
  "cfd": null,
  "fire_points": [],
  "fire_polygons": [],
  "hist_fire": null,
  "is_snow": null,
  "lc": null,
  "prf": null
}
```
And so while I'm not sure it matters much, but should we consider returning `null` rather than an empty list? What do you think?

charparr:
This PR also chips away at XREF #137

BobTorgerson:
We actually do want the code to return empty lists here rather than null to match the code that is expecting to be able to iterate through a list-like object. We could modify the fire map code, but I think it doesn't hurt to leave it like it is and return the empty lists.

charparr (approved):

PR #463, opened 2024-10-03
Always alphabetizes the list of category names on index page

### LINKED ISSUE #455: Always alphabetize the list of endpoints on the home page

BobTorgerson:
This PR moves the location where the service categories are defined so that we can use Python to sort the list by the category name. This list of tuples containing the category name and URL path are passed in sorted order to the index.html file to be rendered.

To test this, you can randomize this list of tuples in the application.py file: https://github.com/ua-snap/data-api/blob/621dbca7adb19ccd1bace2777298dfb39b145dbf/application.py#L21-L39

Closes #455


templates/index.html
```diff
@@ -18,55 +18,56 @@ <h3>Introduction</h3>
   <a href="https://github.com/ua-snap/data-api">GitHub</a>. Instructions on how
   to use each data endpoint is given below.
 </p>
+<!-- 
+The service categories are now defined in the application.py in the get_service_categories function. 
+They are alphebetized by category name and passed to this file.
```
charparr, line 23:
```suggestion
They are alphabetized by category name and passed to this file.
```
Thanks for leaving this comment here, I found it helpful 👍

charparr (approved):
Index page looks good!

See my typo suggestion in the comment and then go ahead and merge. I tested out randomizing the order of the items in application.py and confirm that it had no impact on the rendered template. The index page retained an alphabetized ordering of the endpoints.

PR #465, opened 2024-10-04
Fixes CSV download filenames for all modern browsers

### LINKED ISSUE #454: CSV File Names Incorrectly Formatted with Underscores and Encoding Issues
Our API is generating CSV file names with unwanted leading and trailing underscores, such as `_Precipitation for Utqiaġvik (Barrow).csv_` but there seem to be differences between different platforms and/or browsers.

Example URL: https://earthmaps.io/precipitation/frequency/point/71.2905/-156.789?format=csv&community=AK418

On Charlie-Firefox (MacOS 13.2) the file name is rendered as `_Precipitation for Utqiaġvik (Barrow).csv_` but on Charlie-Safari it is `Future Projections of Precipitation for Utqia%C4%A1vik%20%28Barrow%29.csv` so perhaps there URL encoding issues.
 
Bruce-Firefox on MacOS version 14.3.1 yields no stray underscores but the filename encoding fails like in the prior example.

The expected behavior is that the file name should not have leading or trailing underscores and that the place names should be correctly encoded and displayed across all platforms.

BobTorgerson:
This PR fixes the issue of the filename structure of the download CSVs from the API on different browsers. For example, in Firefox, we were getting: _Precipitation for Utqiaġvik (Barrow).csv_ while in Sarari, we were seeing the HTML entities showing up in the filename like this: Precipitation for Utqia%C4%A1vik%20%28Barrow%29.csv

I modified the code to remove the unneeded double quotes in the response header's representation of the filename, and that fixed the issue per this Stack Overflow article: [CSV Underscore Issue](https://stackoverflow.com/questions/22477273/firefox-and-chrome-appending-underscore-before-and-after-file-name-while-interne)

Additionally, I found that if I remove the second properties["filename"] in this code: https://github.com/ua-snap/data-api/blob/01510c3cae46d2e588f625fce96fa2f14b3e0c03/csv_functions.py#L228-L234 that the file name adds the HTML entities back into the file name for Safari. Thus, it is a required portion of this code to work across all of the browsers.

This also has Python Black format changes as we have not run the linter on that code in a while.

Closes #454

### DIFF
--- a/csv_functions.py
+++ b/csv_functions.py
@@ -138,7 +138,9 @@ def csv_metadata(place_name=None, place_id=None, place_type=None, lat=None, lon=
         if place_type == "huc12":
             metadata += "# Corresponding HUC12 code: " + place_id + "\n"
     elif place_name is None and lat is None and lon is None:
-        metadata += "All communities Alaska\n" # this covers the demographic request for "all"
+        metadata += (
+            "All communities Alaska\n"  # this covers the demographic request for "all"
+        )
     elif place_type == "community":
         metadata += place_name + "\n"
     else:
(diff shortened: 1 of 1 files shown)


charparr (approved):
This is awesome! I'm getting nice file names now like `Demographic Data - for Anchorage (Dgheyaytnu)` and `Climate Protection from Spruce Beetles for 64.88, -147.29.csv` instead of wacky ones littered with underscores and URL-encodings. The new CSVs are opening in the "Numbers" app without it yelling at me.

PR #466, opened 2024-10-04
Removes the Obu and Jorg fields from the returned permafrost CSVs

### LINKED ISSUE #332: Drop the non-dataframe stuff from CSV version of the Permafrost route
We don't want the obu/jorgenson stuff in there, omit it.  we don't want to bake heterogenous data structures into a single CSV because it doesn't make sense in terms of a user doing analysis on the data.

### LINKED ISSUE #438: Drop Obu, Jorg etc fragments from CSV
Rows 616-618.  columns O-R, remove and update the header of the CSV as needed.

BobTorgerson:
This PR removes the three lines of code that added the Obu and Jorg fields to the returned CSVs for the permafrost endpoint that are no longer referenced in the header information of the CSV.

An example endpoint: http://localhost:5000/permafrost/point/65.0628/-146.56?format=csv

Closes #438 
Closes #332

### DIFF
--- a/csv_functions.py
+++ b/csv_functions.py
@@ -504,9 +504,6 @@ def permafrost_csv(data, source_metadata):
                 "talikthickness",
             ],
         },
-        "jorg": {"coords": ["source"], "values": ["ice", "pfx"]},
-        "obu_magt": {"coords": ["source"], "values": ["year", "depth", "temp"]},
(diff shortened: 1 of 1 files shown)


charparr (approved):
The new CSVs look great.

PR #467, opened 2024-10-04
Removes Geology from the API

### LINKED ISSUE #436: Remove Geology from API entirely
We're done with geology.  Remove.

BobTorgerson:
This PR removes Geology as an available resource on the API entirely, removing any references to it, including in the README document. I tested the new /eds/all endpoint that contained Geology previously against a local copy of the ArcticEDS and the API continues to give the correct results to the EDS. We didn't have any references to Geology anywhere in the EDS.

Closes #436


charparr (approved):
See ya later 🪨 endpoint, you had a good run!

This PR effectively removes all references to 'geology' and I've verified the that `eds/all` route still functions as expected. Both Bob and I have queried the text of the repo for references to "geology" and can verify they are removed.

I've also verified that the `eds/all` endpoint still works as expected, e.g., http://127.0.0.1:5000/eds/all/65.0628/-146.1627

PR #468, opened 2024-10-07
Additional CSV fix for latitude and longitude filenames

BobTorgerson:
This PR fixes the file names with latitude and longitude and doesn't have a community ID name. The previous fix for the special characters working the same way in all of the browsers removed quotes that were preventing the comma in the latitude and longitude file name from breaking the write_csv function. 

This removes the comma between the latitude and longitude and all of the special characters work across all of the major browsers still.

### DIFF
--- a/csv_functions.py
+++ b/csv_functions.py
@@ -110,7 +110,7 @@ def create_csv(
         elif endpoint == "demographics":
             filename += quote("All communities in Alaska")
         else:
-            filename += lat + ", " + lon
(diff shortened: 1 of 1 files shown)


charparr (approved):
We got it now! I'm testing routes like this http://127.0.0.1:5000/degree_days/below_zero/65.0628/-146.1627?format=csv

and getting satisfactory CSV names like `Degree Days Below Zero for 65.0628 -146.1627.csv`

PR #464, opened 2024-10-03
Modifies EDS endpoint for Wet Days Per Year with checks for error conditions being returned

BobTorgerson:
This PR adds the missing checks for returned error tuples in the projected data results and rendering them if there is an issue getting data back from Rasdaman.

This will fix the case where the network cuts out or the Rasdaman server returns any of the error codes we expect to need to render for the /eds/wet_days_per_year/point/

Working endpoint: http://localhost:5000/eds/wet_days_per_year/point/62.15/-155.65
Failing endpoint showing error condition: http://localhost:5000/eds/wet_days_per_year/point/62.15/-115.65

Non-working endpoint on production site: https://earthmaps.io/eds/wet_days_per_year/point/62.15/-115.65

NOTE: The error case in #459 could only occur previous when the data is returned correctly for the historical data and incorrectly for any of the projected data which would be a 404 or 500 error.

Closes #459

PR #473, opened 2024-10-15
Fixes tuple return error in ALFRESCO endpoint

### LINKED ISSUE #135: Some ALFRESCO local endpoints are attempting to assign to a tuple
- http://localhost:5000/alfresco/flammability/local/52.938/-168.868
- http://localhost:5000/alfresco/flammability/local/53.267/-168.217

This code [on line 573 ](https://github.com/ua-snap/data-api/blob/f9130b2f43b19f5c9a8df9fb898ad3396b63158e/routes/alfresco.py#L573) is returning as a tuple in some cases and resulting in this error.

Joshdpaul:
Closes #135 

This is a similar problem to the one fixed in PR #464 , and I just copied the solution from there.

To test:

Fire up the API and make sure these endpoints render a template instead of a tuple error:

http://localhost:5000/alfresco/flammability/local/52.938/-168.868
http://localhost:5000/alfresco/flammability/local/53.267/-168.217

### DIFF
--- a/routes/alfresco.py
+++ b/routes/alfresco.py
@@ -275,7 +275,12 @@ def run_fetch_alf_local_data(var_ep, lat, lon):
     if validation == 400:
         return render_template("400/bad_request.html"), 400
     if validation == 422:
-        return render_template("422/invalid_latlon.html", west_bbox=WEST_BBOX, east_bbox=EAST_BBOX), 422
+        return (
+            render_template(
+                "422/invalid_latlon.html", west_bbox=WEST_BBOX, east_bbox=EAST_BBOX
+            ),
+            422,
+        )
 
     # Requests for HUC12s that intersect
     huc12_features = asyncio.run(
--- a/routes/alfresco.py
+++ b/routes/alfresco.py
@@ -275,12 +275,7 @@ def run_fetch_alf_local_data(var_ep, lat, lon):
     if validation == 400:
         return render_template("400/bad_request.html"), 400
     if validation == 422:
-        return (
-            render_template(
-                "422/invalid_latlon.html", west_bbox=WEST_BBOX, east_bbox=EAST_BBOX
-            ),
-            422,
-        )
+        return render_template("422/invalid_latlon.html", west_bbox=WEST_BBOX, east_bbox=EAST_BBOX), 422
 
     # Requests for HUC12s that intersect
     huc12_features = asyncio.run(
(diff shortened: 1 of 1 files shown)


routes/alfresco.py
```diff
@@ -275,12 +275,7 @@ def run_fetch_alf_local_data(var_ep, lat, lon):
     if validation == 400:
         return render_template("400/bad_request.html"), 400
     if validation == 422:
-        return (
-            render_template(
-                "422/invalid_latlon.html", west_bbox=WEST_BBOX, east_bbox=EAST_BBOX
-            ),
-            422,
-        )
+        return render_template("422/invalid_latlon.html", west_bbox=WEST_BBOX, east_bbox=EAST_BBOX), 422
```
BobTorgerson, line 278: My Python Black formatting converted this back to the original format that was changed here when saving this file locally. I think our Python Black configurations must be off. @charparr Are you seeing the same thing when you run the formatter on this file?

BobTorgerson (changes requested):
This change works and prevents error conditions from slipping through and causing 500 errors further down the chain of processing. However, I did note in my review that the change to line 278 was reverted when I ran Python Black on it, which means that our configurations are different. We should figure out what difference your IDE has compared to mine if Charlie confirms that his Python Black configuration also reverts that change.

routes/alfresco.py
charparr: @BobTorgerson same, here is my diff when I save that file (I `black` on save of every python file automatically in VS Code):
```
-        return render_template("422/invalid_latlon.html", west_bbox=WEST_BBOX, east_bbox=EAST_BBOX), 422
+        return (
+            render_template(
+                "422/invalid_latlon.html", west_bbox=WEST_BBOX, east_bbox=EAST_BBOX
+            ),
+            422,
+        )
```

The thing about Black is that there shouldn't even be a configuration: it is a strict, opinionated formatter. So long as we are running black on save or pre-commit then we shouldn't have any differences across our linting.
@Joshdpaul what is your setup?

Joshdpaul: Ah yes, I've had this issue before with getting the remote connection to actually do the black format on save! Sometimes hard to tell when it's working or not.

I tried reinstalling the extension and re-saving the file, and line 278 now looks like your code above. Black adds parentheses around the return tuple, which is interesting...I originally edited that to be consistent with the 400 return line above, but if that's not necessary than we can go with the black formatting.

charparr: Yeah, one pitfall I have found is that every time I connect to a new remote I need to re-install extensions. And so sometimes I can save a file, anticipating it will black, and then it isn't there. Two tell-tale black signals are that double quotes are nearly everywhere, rather than single quotes, and that all operators are bookended by white space, like `a=3` would be black formatted to `a = 3`.

BobTorgerson: Yeah, it's no big deal! We all miss the formatting and it's honestly not a problem at all! Even if it slips through the cracks (which it most certainly does), it will be caught later by someone else and the correct formatting will be applied. This was an easy PR to identify that as being a potential formatting difference, so I wanted to ensure we weren't thrashing with config changes. I approve this change now!

PR #479, opened 2024-10-28
Replaces all unnecessary star imports for functions across the API and changes fetch_data_api function name

### LINKED ISSUE #349: Rename `fetch_data_api` to indicate GS specific function
This function is only used for WFS / WMS queries to Geoserver (though the actual backend is flexible). I'd like to see this function renamed because it isn't very specific (this function is probably one of the oldest in the repo, which is why it has such a general / confusing name).

https://github.com/ua-snap/data-api/blob/cb282a5702c04ae00907c6ba272e332d97493fe6/fetch_data.py#L69

### LINKED ISSUE #401: Remove star imports from each route
Many of our routes have star imports, e.g., `from fetch_data import *` and while this is handy for dev work, it is polluting our namespace and is a bit of a known Python anti-pattern. We should import just the functions we need: this will also help us ID and reduce the number of duplicate functions and code over time.

SO reference thread: https://stackoverflow.com/questions/2386714/why-is-import-bad

XREF #399

BobTorgerson:
This PR replaces all of the unnecessary star imports in the API and explicitly calls the functions that it needs to operate in each route and functionality. 

To test, ensure that all of the endpoints are still operational after this change as this touches a number of different routes and files that contain functions used across the API. 

Closes #349 
Closes #401

### DIFF
--- a/application.py
+++ b/application.py
@@ -3,7 +3,7 @@ from flask import Flask, render_template, send_from_directory
 from flask_cors import CORS
 from config import SITE_OFFLINE
 
-from routes import *
+from routes import routes, request
 
 # Elastic Beanstalk wants `application` to be present.
 application = app = Flask(__name__)
--- a/csv_functions.py
+++ b/csv_functions.py
@@ -1,8 +1,12 @@
-from flask import request, Response, request
+from flask import request, Response, render_template
+import copy
 import csv
-from postprocessing import *
-from fetch_data import *
-from validate_data import *
+import io
+from urllib.parse import quote
+from postprocessing import nullify_and_prune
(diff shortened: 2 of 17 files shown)


cstephen (approved):
I tested a ton of JSON and CSV links from the Data API documentation, and also the links generated by local instances of Northern Climate Reports, Arctic-EDS, and ARDAC Explorer. Every link worked as expected. 💯

charparr (approved):
I've tried a bunch of endpoints as they are all functioning as expected 💯 thanks for taking the time to go through all these routes and clean up the imports. I think it'll help us down the road and it also makes the code more searchable and readable I think.

PR #491, opened 2024-11-14
New demographics fields + improved CSV 

### LINKED ISSUE #475: also filter data from CSV if population < x
When we choose to not display data based on population (e.g., # of adults < 50) we should also not serve this data in the CSV download.

### LINKED ISSUE #476: Add field descriptions for demographic data?
Add a "horizontal" row lined up with the data that gives the definition of the field.  Would need to be comma/special-char escaped.

### LINKED ISSUE #482: Add data descriptions to demographics CSV output
Add more descriptive metadata and text to the CSV output of the demographics data.

### LINKED ISSUE #487: change field order in demographics data CSV output
Make it match the field order as displayed in NCR

### LINKED ISSUE #489: limit demographics data returns based on population threshold
Whatever population threshold we are using to limit data shown in NCR, we need to apply this same threshold in the API. 

For the CSV: NCR invites users to download the CSV, so that seems like it should have the population threshold applied. 
For the JSON: should this data access pathway also have a threshold applied?

### LINKED ISSUE #490: Update demographics shapefile field crosswalk with new PLACES / SDOH data fields
If / when we update the demographics shapefile on GeoServer, it will have some new fields (truncated to 10 characters) that need to be added to the crosswalk in `luts.py`

Joshdpaul:
This PR fetches the new fields added to the demographics data [during this PR](https://github.com/ua-snap/epa-justice/pull/15), limits API returns to communities with >=50 people,  and improves the CSV output by tranposing the table, adding more detailed descrioptions of the data elements, and re-ordering the fields to match the display order in NCR. 

closes #490 
closes #489 
closes #487 
closes #482
closes #476 
closes #475 

To test:
- set the `API_GS_BASE_URL` environmental variable to the development server like so:
```export API_GS_BASE_URL=https://gs.earthmaps.io/geoserver/```
- fire up the API and request demographics data (including CSVs) for some communities. Communities with <50 people should return a basic JSON error message. For example, try Fairbanks and Nabesna.
```
# Fairbanks should return data as JSON and CSV
http://127.0.0.1:5000/demographics/AK124
http://127.0.0.1:5000/demographics/AK124?format=csv

# Nabesna should not return data as JSON or CSV
http://127.0.0.1:5000/demographics/AK256
http://127.0.0.1:5000/demographics/AK256?format=csv
```

The descriptive text in the CSV output has grown considerably here, and so could use some careful proofreading to make sure the sentences are sane and reference the actual datasets that the particular line of data came from. Crosschecking with the `README.md` would be much appreciated!

**Note:** In order to use `gs.earthmaps.io` in this branch, I had to bypass SSL certificate verification in the route-specific `demographics.validate_community_id()` function and also in more widely used `fetch_data.make_get_request()` function. I didn't know how else to get this to work, but I imagine we need to change this in production. Any thoughts on how & when is best to re-introduce the SSL verification would be welcomed!

### CODE REVIEW

File: luts.py [hunk 1]
```diff
@@ -133,5 +134,142 @@
...
+    "pct_foodstamps_high": "pct_foodstamps_high is the upper bound of the 95% confidence interval for percentage of adults aged >=18 years that received food stamps in the past 12 months as reported in the CDC PLACES survey for 2024",
+    "pct_w_disability": "pct_w_disability is the percentage of population with a disability as reported in the Census American Community Survey 5-year survey for years 2018-2022",
+    "moe_pct_w_disability": "moe_pct_w_disability is the margin of error for percentage of population with a disability as reported in the Census American Community Survey 5-year survey for years 2018-2022",
+    "pct_insured": "pct_insured is the percentage of population with health insurance as reported in the Census American Community Survey 5-year survey for years 2018-2022",
+    "moe_pct_insured": "moe_pct_insured is the margin of error percentage of population with health insurance as reported in the Census American Community Survey 5-year survey for years 2018-2022",
```
cstephen, line 203:
```suggestion
    "moe_pct_insured": "moe_pct_insured is the margin of error for percentage of population with health insurance as reported in the Census American Community Survey 5-year survey for years 2018-2022",
```

luts.py
```diff
@@ -133,5 +134,142 @@
...
+    "pct_uninsured": "pct_uninsured is the percentage of population without health insurance as reported in the Census American Community Survey 5-year survey for years 2018-2022",
+    "moe_pct_uninsured": "moe_pct_uninsured is the margin of error percentage of population without health insurance as reported in the Census American Community Survey 5-year survey for years 2018-2022",
```
cstephen, line 205:
```suggestion
    "moe_pct_uninsured": "moe_pct_uninsured is the margin of error for percentage of population without health insurance as reported in the Census American Community Survey 5-year survey for years 2018-2022",
```

cstephen (changes requested):
Things are looking really good in this PR! I gave it a thorough review and have a handful of suggestions ranging from minor to nitpicky:

- The endpoint is currently not returning data if the population is under 50, but to be more precise (and persnickety, admittedly), we are supposed to not return data if the _adult_ population is under 50. We can do that by subtracting the number of people under 18 from the total population like we did in NCR here: https://github.com/ua-snap/northern-climate-reports/blob/55c92c7ec20ac76c59963d07f87d476682e3d880/components/reports/demographics/DemographicsReport.vue#L110-L117
- To be consistent with our other endpoint behavior, IMO the error returned when the population is too low (for both http://127.0.0.1:5000/demographics/AK256 and http://127.0.0.1:5000/demographics/AK256?format=csv) should use an HTML template. See, for example, how these endpoints return the error embedded in an HTML template regardless of whether you request the response as JSON or CSV: http://localhost:5000/temperature/65.0628/-146.1627/1899/2060 and http://localhost:5000/temperature/65.0628/-146.1627/1899/2060?format=csv
I think we still want a "community population <50 people, no data returned" message, though, so we may need to create a new HTML template for this error message.
- I noticed when comparing the endpoint fields to the epa-justice repo README that the "pct_hcost" field is missing from both the JSON output and CSVs. Not sure if this is intentional or not.
- The "comment" row at the bottom of the CSV file seems a little out of place, and we may want to move this near the top of the CSV file as a metadata line.
- (Very nitpicky), but I think it'd be better if the "Alaska" column is to the left of the "United States" column in the CSV, so the columns are in this order (for example): Fairbanks / Alaska / United States.
- I've added a couple PR code comments for a couple instances of missing words in the field descriptions.

Also, I experimented with adding SSL verification back and didn't run into any issues, surprisingly! So, we may be able to add SSL verification back unless it's still not working on your side, but we can figure out what's up with that if so.

Joshdpaul:
@cstephen Thank you for the detailed review! Great job catching all those little things, they are important!

I was able to address all your comments, and took a stab at a 403 template. Not sure if that's the most appropriate code, or if the message should be something else, but we can adjust that pretty easily in future work.

I am still unable to make requests without turning off SSL verification. I get this error:

```
SSLError
requests.exceptions.SSLError: HTTPSConnectionPool(host='gs.earthmaps.io', port=443): Max retries exceeded with url: /geoserver/wfs?service=WFS&version=2.0.0&request=GetFeature&typeName=demographics:demographics&outputFormat=application/json&propertyName=(id) (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1135)')))
```

routes/demographics.py
```diff
@@ -85,15 +142,33 @@ def get_data_for_community(community):
         for field in fields:
             reformatted_results[c][field] = results[c][field]
 
+    # apply population threshold
+    if (
+        reformatted_results[community]["total_population"]
+        - reformatted_results[community]["pct_under_18"]
+    ) < 50:
+        return render_template("/403/pop_under_50.html"), 403
```
cstephen, line 150:
```suggestion
    totalPopulation = reformatted_results[community]["total_population"]
    percentUnder18 = reformatted_results[community]["pct_under_18"]
    populationUnder18 = totalPopulation * (percentUnder18 / 100)
    adultPopulation = round(totalPopulation - populationUnder18)

    # apply population threshold
    if adultPopulation < 50:
        return render_template("/403/pop_under_50.html"), 403
```

Looks like you did the same thing I did at first when doing this on the NCR side (subtracting the raw percentage instead of the population represented by the percentage), which is why this jumped out at me. Easy fix though!

cstephen (changes requested):
@Joshdpaul, just one more code change (see PR code suggestion) and this PR looks good to me! Also, I researched HTTP status codes and 403 does seem like a very fitting code for this case, so sounds good to me 👍

routes/demographics.py
Joshdpaul: Ah yes! Good catch! and +1 for lowerCamelCase 🐫  😃

cstephen: Dang, actually I shouldn't have been using lowerCamelCase here! My head was full of JavaScript. lowerCamelCase is considered not-very-Python. Feel free to change these to snake_case!!

Joshdpaul: Doesn't bother me really, the code works either way! What do you think is best?

cstephen: We should probably change these variables to snake_case so as not to incur the wrath of others 😅 Sorry for the mess!

Joshdpaul: done! 🐍

cstephen (approved):

cstephen (approved):

PR #494, opened 2024-11-19
Validate CMIP6 lat/lons against CMIP6-specific BBOX

### LINKED ISSUE #493: Fix BBOX validation for CMIP6 monthly and CMIP6 indicator endpoints
Looks like requests to the CMIP6 monthly and CMIP6 indicator endpoints are validating against the default BBOXs for the API instead of allowing any point north of ~50 latitude. I hadn't caught this before because I'd gotten into the habit of testing against communities.

cstephen:
Closes #493.

This PR validates lat/lon coordinates for CMIP6-based endpoints against a CMIP6-specific BBOX, which includes everything above ~49.95 latitude. Without this change, CMIP6 endpoints were being validated against the same Alaska-centric East/West BBOXes as the other endpoints, which is just a small subset of the CMIP6 extent.

To test, verify that the following example URLs return data:

http://localhost:5000/indicators/cmip6/point/85/175
http://localhost:5000/cmip6/point/85/175?vars=tas,pr

And also verify that the following two URLs return an error page describing the valid CMIP6 BBOX:

http://localhost:5000/indicators/cmip6/point/30/175
http://localhost:5000/cmip6/point/30/175

@charparr, I'm tagging you for this review since (if memory serves) you were working on an overhaul of the BBOX validation error page recently, and I want to make sure this doesn't conflict with your work. Feel free to modify this PR as needed if so!

### DIFF
--- a/config.py
+++ b/config.py
@@ -9,7 +9,7 @@ RAS_BASE_URL = os.getenv("API_RAS_BASE_URL") or "https://apollo.snap.uaf.edu/ras
 WEST_BBOX = [-180, 51.3492, -122.8098, 71.3694]
 EAST_BBOX = [172.4201, 51.3492, 180, 71.3694]
 SEAICE_BBOX = [-180, 30.98, 180, 90]
-INDICATORS_BBOX = [0, 49.94, 359.37, 90]
+CMIP6_BBOX = [-180, 49.95, 180, 90]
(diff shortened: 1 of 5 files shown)


cstephen:
Closing this PR without merging since it conflicts with the work merged in PR #495 (as expected and noted in my previous comment), and the final step of removing the deprecated WEST_BBOX and EAST_BBOX check will be implemented as a new, simpler PR, later today.

Issue #481, opened 2024-11-07
Modify the `get_dim_encodings` function to account for differences in the default DescribeCoverage request + response in Rasdaman.


charparr:
This issue is motivated by a bug encountered while developing new API endpoints. The STR to reproduce this bug are to set your the Rasdaman env var to:

`export API_RAS_BASE_URL=https://datacubes.earthmaps.io/rasdaman/`

and the observe this error which (a red herring, as the coverages listed as missing do in fact exist):
```
Warning: Coverage 'crrel_gipl_outputs' is missing from the Rasdaman server https://datacubes.earthmaps.io/rasdaman/ you are using.
Warning: Coverage 'alfresco_relative_flammability_30yr' is missing from the Rasdaman server https://datacubes.earthmaps.io/rasdaman/ you are using.
Warning: Coverage 'alfresco_vegetation_type_percentage' is missing from the Rasdaman server https://datacubes.earthmaps.io/rasdaman/ you are using.
Warning: Coverage 'heating_degree_days_Fdays' is missing from the Rasdaman server https://datacubes.earthmaps.io/rasdaman/ you are using.
Traceback (most recent call last):
  File "/Users/cparr/.local/share/virtualenvs/data-api-_zt2_2sd/bin/flask", line 8, in <module>
    sys.exit(main())
  File "/Users/cparr/.local/share/virtualenvs/data-api-_zt2_2sd/lib/python3.9/site-packages/flask/cli.py", line 1064, in main
    cli.main()
  File "/Users/cparr/.local/share/virtualenvs/data-api-_zt2_2sd/lib/python3.9/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/Users/cparr/.local/share/virtualenvs/data-api-_zt2_2sd/lib/python3.9/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/cparr/.local/share/virtualenvs/data-api-_zt2_2sd/lib/python3.9/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/cparr/.local/share/virtualenvs/data-api-_zt2_2sd/lib/python3.9/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
[...]

PR #495, opened 2024-11-25
Landfastice update

### LINKED ISSUE #220: Landfast Sea Ice: Consider not returning null values
Landfast ice currently returns a `null` value mixed into the time series alongside `1` values that indicate the presence of landfast sea ice. The endpoint should probably implement a data return of 1s and 0s (or even true and false) instead of 1s and nulls to better match the general motif we've created in this Data API. We're expecting a significant re-processing of this dataset - so revisit this ticket when that happens. 

Consider a scheme like: 
1: landfast sea ice presence
0: landfast sea ice absence, but possible for that location
255: definitely a terrestrial location - call me NoData
Some other value (254?): Coastal location, but could be terrestrial we can't rule out that there could be landfast sea ice data here (or just make these zero also)
Some other value (254?): Open ocean, landfast ice never happens here (e.g., this is always pack ice). Is there some known distance from the coast at which this condition occurs?

It'd be great to get a coastal mask from Hajo and Andy - especially if it was the one used to generate these data in the first place!

### LINKED ISSUE #221: Marine Datasets may need to override the BBOX
The current upper bound of `71.3694` might truncate some marine datasets like landfast sea ice. These data may need to override that config or import something else, e.g., `MARINE_BBOX_NORTH` or something like that. We want to avoid an unneeded 422 when there is actually data.

charparr:
This PR basically swaps the backend for the landfast sea ice endpoint. The endpoint is functionally quite similar, but the geographic and temporal ranges are extended. The possible set of returned values is also a bit different. Check the refreshed template for details. Here's what to look for in this review (some of this was foreshadowed in our tech gardening / show+tell chat):

#### A new implementation of the DescribeCoverage request

* Checkout the new function `generate_wcps_describe_coverage_str` in the `generate_requests` module. We're forming a request to describe the coverage using the WCPS `describe()` operator because we can get a response that is a well-formatted JSON that is far easier to parse than the prior XML/string babble. We'll see how this is used later on.
* Next, look at the new `generate_describe_coverage_url` function in the `generate_urls` module - this just inserts the previous function's return into a URL so we can make the request.
* Now, look at the new `describe_via_wcps` function in the `fetch_data` module  - this function is responsible for firing off the awaitable request and receiving the response.
* None of these three functions are that much different that what we've been doing already, and some of the other functions could probably have been extended to accomplish the goal of describing coverages via the WCPS `describe()` operator - but I'm personally in favor of having functions that do one thing and one thing only when possible: this chain of execution where `describe_via_wcps()` calls `generate_wcps_describe_coverage_str()` and `generate_describe_coverage_url()`
has the sole purpose of fetching a coverage description in JSON format.

#### A new implementation of lat-lon request validation

The JSON-ified coverage description makes it so much easier to determine what spatial bounding box is valid for a coverage. With such bounding boxes, we can validate requests more precisely on a coverage-by-coverage basis, and can leverage the self-describing nature of OGC Coverages. Our current practice is to prescribe ad hoc bounding box prescriptions in `config` that are decoupled from the coverage they are referencing. So, there are **a few new functions in the `validate_request` module to know about!**

* First, look at the `get_x_y_axes()` function - this will just ID and return which axes of the coverage have the spatial (i.e. X and Y) information.
* Second, see the `construct_latlon_bbox_from_coverage_bounds()` function that constructs a lat-lon bounding box from the metadata returned by `describe_via_wcps`. These bounding boxes are just like those prescribed in our `config.py` and used in our existing validation scheme. Note that this function just returns the bounding box, and doesn't do any validation.
* Third, take a look at the `validate_latlon_in_bboxes()` function that iterates through a list of bounding boxes and so long as the given lat and lon are within **at least one of those boxes** a `True` value is returned, and otherwise a `422` code is thrown. A list of bounding boxes is implemented because the landfast sea ice data is split between two coverages, and we need to know if a query lands in either one.
* Next, look at `validate_xy_in_coverage_extent()` - at first glance this function might seem a bit redundant with the previous function. But this function is specific: it returns a boolean based on whether or not an x-y coordinate is within a single coverage's bounding box. One way to distinguish this function from the previous could be to think of it this way: This asks "Am in the bounding box for this specific coverage?" while the previous function asks "Am in the possible set of bounding boxes encompassed by this endpoint?"
* Finally, there is a function called `latlon_is_numeric_and_in_geodetic_range()` that does the “bad request” part of the validation. This function splits out part of the existing functionality of the `validate_lat_lon` function, because the `validate_lat_lon` function actually doing two things: checking the bbox against the config prescription, and ensuring that the lat-lon is numeric and on planet Earth. Observe the repeated code between `validate_lat_lon` and `validate_seaice_lat_lon`. The new `latlon_is_numeric_and_in_geodetic_range()` function is a step toward DRY-ing the request validation, and is the first function in the validation tree for the updated landfast sea ice endpoint. Also, just to avoid confusion, the `validate_seaice_lat_lon` is **not** for the landfast sea ice endpoint considered in this PR.

#### Whew, OK, now to test the endpoint itself:

* First, you need to modify the imports in `routes/__init__.py` to look like this (for details XREF #481):
```python
# from .fire import *
# from .permafrost import *
# from .seaice import *
# from .taspr import *
# from .physiography import *
# from .boundary import *
# from .vectordata import *
# from .recache import *
# from .elevation import *
# from .alfresco import *
# from .degree_days import *
# from .snow import *

from .landfastice import *

# from .beetles import *
# from .eds import *
# from .wet_days_per_year import *
# from .indicators import *
# from .hydrology import *
# from .demographics import *
# from .cmip6 import *
```

* Point the API at Datacubes aka Zeus:
```sh
export API_RAS_BASE_URL=https://datacubes.earthmaps.io/rasdaman
```

- Now, try some requests that **should not yield a valid data response.**
  - Requests like these should trigger a 400 Bad Request template:
    - http://localhost:5000/landfastice/point/400/-12345/
    - http://localhost:5000/landfastice/point/foo/bar/
  - Requests like these should trigger a 422 and send you to a brand new `invalid_latlon_outside_coverage` template! Check out this template - what do you think? Suggestions? Also, please look at the code for the template itself closely as I had to lean on AI heavily here to figure out the leaflet JS part.
    - http://localhost:5000/landfastice/point/65/-145/
    - http://localhost:5000/landfastice/point/62/-172/
  - Requests like these should trigger a 404 No Data template.
    - http://localhost:5000/landfastice/point/71/-134/
    - http://localhost:5000/landfastice/point/71/-172/
- Now, try some requests that **should yield a valid data response.** Some of these might take a while to complete, so you may need to refresh if it seems like they time out.
  - http://localhost:5000/landfastice/point/70.6123/-149.8869/
  - http://localhost:5000/landfastice/point/71.2926/-156.7956?format=csv
  - And here are some other lat-lons you might try for pixels that are in the ocean, but are near a named community:
```python
Deadhorse: {'latitude': 70.2993, 'longitude': -148.4942}
Kaktovik: {'latitude': 70.1056, 'longitude': -143.6875}
Nuiqsut: {'latitude': 70.3527, 'longitude': -151.1869}
Prudhoe Bay: {'latitude': 70.3028, 'longitude': -148.3779}
Utqiaġvik: {'latitude': 71.2926, 'longitude': -156.7956}
Wainwright: {'latitude': 70.6301, 'longitude': -160.2603}
Candle: {'latitude': 66.0019, 'longitude': -161.933}
Cape Lisburne: {'latitude': 68.8687, 'longitude': -166.2243}
Deering: {'latitude': 66.0759, 'longitude': -162.7157}
Diomede: {'latitude': 65.7769, 'longitude': -168.8938}
Elephant Point: {'latitude': 66.2676, 'longitude': -161.3338}
Kivalina: {'latitude': 67.7282, 'longitude': -164.532}
Kiwalik: {'latitude': 66.0336, 'longitude': -161.8338}
Kotzebue: {'latitude': 66.8995, 'longitude': -162.5982}
Point Hope: {'latitude': 68.3447, 'longitude': -166.7331}
Point Lay: {'latitude': 69.7577, 'longitude': -163.052}
Shishmaref: {'latitude': 66.2569, 'longitude': -166.0729}
Tin City: {'latitude': 65.5707, 'longitude': -168.048}
Wales: {'latitude': 65.6087, 'longitude': -168.0923}
```

Closes #220 
Closes #221

### DIFF
--- a/csv_functions.py
+++ b/csv_functions.py
@@ -77,8 +77,8 @@ def create_csv(
         properties = gipl_csv(data, endpoint)
     elif endpoint in ["ncar12km_indicators"]:
         properties = ncar12km_indicators_csv(data)
-    elif endpoint == "landfastice":
-        properties = landfastice_csv(data)
+    elif endpoint == "landfast_sea_ice":
+        properties = landfast_sea_ice_csv(data)
     elif endpoint == "permafrost":
         properties = permafrost_csv(data, source_metadata)
     elif endpoint.startswith("places_"):
@@ -558,18 +558,16 @@ def ncar12km_indicators_csv(data):
     }
 
 
-def landfastice_csv(data):
+def landfast_sea_ice_csv(data):
     # Reformat data to nesting structure expected by other CSV functions.
     for key, value in data.items():
-        data[key] = {"status": value}
+        data[key] = {"value": value}
     coords = ["date"]
-    values = ["status"]
+    values = ["value"]
     fieldnames = coords + values
     csv_dicts = build_csv_dicts(data, fieldnames, values=values)
-    metadata = (
-        "# Landfast Ice Status: A 0 value indicates absence and 1 indicates presence.\n"
-    )
-    filename_data_name = "Landfast Ice Extent"
+    metadata = "# Landfast Sea Ice Value Key: 0: Open ocean or non-landfast sea ice; 128: Land; 255: Landfast Sea Ice\n"
+    filename_data_name = "Landfast Sea Ice"
 
     return {
         "csv_dicts": csv_dicts,
--- a/fetch_data.py
+++ b/fetch_data.py
@@ -20,13 +20,18 @@ from aiohttp import ClientSession
 from flask import current_app as app
 from rasterstats import zonal_stats
 from config import RAS_BASE_URL
-from generate_requests import generate_wcs_getcov_str, generate_netcdf_wcs_getcov_str
+from generate_requests import (
+    generate_wcs_getcov_str,
+    generate_netcdf_wcs_getcov_str,
+    generate_wcps_describe_coverage_str,
+)
 from generate_urls import (
     generate_wcs_query_url,
     generate_base_wms_url,
     generate_base_wfs_url,
     generate_wms_and_wfs_query_urls,
     generate_wfs_places_url,
+    generate_describe_coverage_url,
 )
 
 
@@ -586,3 +591,18 @@ def replace_nans(json_str):
     # This is to prevent matches against strings that contain 'nan' within them.
     json_str = re.sub(r"(?<=[,\[\]])nan(?=[,\[\]])", "-9999", json_str)
     return json_str
+
+
+async def describe_via_wcps(cov_id):
+    """Get the metadata in JSON format via a WCPS describe() query request.
+
+    Args:
+        cov_id (str): rasdaman coverage ID
+
+    Returns:
+        json_description (dict): coverage description in JSON format
+    """
+    req_str = generate_wcps_describe_coverage_str(cov_id)
+    req_url = generate_describe_coverage_url(req_str)
+    json_description = await fetch_data([req_url])
+    return json_description
--- a/generate_requests.py
+++ b/generate_requests.py
@@ -196,3 +196,17 @@ def generate_netcdf_average_wcps_str(bbox_bounds, generate_average_wcps_str_kwar
         **generate_average_wcps_str_kwargs,
     )
     return netcdf_avg_wcps_str
+
+
+def generate_wcps_describe_coverage_str(cov_id):
+    """Generate a WCPS DescribeCoverage request for a given coverage.
+
+    The describe() operation returns a description of the coverage equivalent to a WCS DescribeCoverage request, and the output adheres to the same WCS schema, but we're able to get it in JSON format instead of only XML, GML, or string.
+
+    Args:
+        cov_id (str): rasdaman coverage ID
+    Returns:
+        describe_coverage_str (str): encoded WCPS string fragment to append to a query URL
+    """
+    query_str = f'for $c in ({cov_id}) return describe($c, "application/json", "outputType=GeneralGridCoverage")'
+    return quote(query_str)
--- a/generate_urls.py
+++ b/generate_urls.py
@@ -98,3 +98,14 @@ def generate_wms_and_wfs_query_urls(wms, wms_base, wfs, wfs_base):
     for veclyr in wfs:
         urls.append(wfs_base.format(veclyr, wfs[veclyr]))
     return urls
+
+
+def generate_describe_coverage_url(describe_coverage_str):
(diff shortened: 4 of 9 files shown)


cstephen (changes requested):
I read through the PR description and tried out the example links (and a few random ones), and everything's looking almost perfect to me. Pulling the coverage bounding box from Rasdaman and then using that to populate the error template w/ the Leaflet map visualization is a game changer, and a huge step up from our previous implementation.

I looked over the Leaflet code and it looks great to me. The way that the code calculates the `biggest_bounds` from the min/max values of the individual bounding boxes, and then fits the map to `biggest_bounds` is a very smart approach.

I noticed a couple oddities while poking around:

- http://localhost:5000/landfastice/point/71/-134/ returns a HTTP 500 error page on my end, not an HTTP 404 as mentioned in the PR description.
- The CSV from http://localhost:5000/landfastice/point/71.2926/-156.7956?format=csv has lots of blank values in it. I assumed the 0 values were getting changed to blank values or something, but other CSV endpoints have both 255 and 0 values (and no blank cells as far as I can tell): http://localhost:5000/landfastice/point/68.8687/-166.2243/?format=csv I assume the blank values are a problem, but if not, we may want to explain them in the CSV metadata.

Otherwise, this looks mergable to me!

charparr:
OK thanks for the review Craig! I think I've got it figure out: what I've done is break out the `tolerance` parameter into four directional components to prevent over-padding the bounding boxes and so that we only pad in the direction of where the distortion at the edge of the projected reference system will occur. I think the unexpected 500 we had earlier was because the coordinate wasn't landing in either of the projected bounding boxes.

I've also deployed the "nullify_with_max_intensity" function to prune the nodata values that were causing blanks in the CSV responses. I had to refresh my memory: this function will prune keys even when the same-level sibling key has data, which is the behavior we want here, vs. the less aggressive pruning that was in place prior.

charparr:
@cstephen after chatting with @brucecrevensten I'm opting for the self-merge here. I'm reasonably confident that I addressed your comments and we can always iterate more upon this later. Onward!

PR #496, opened 2024-12-06
Demographics and health endpoint and CSV updates

cstephen:
This is a draft PR to implement the latest round of demographics & health changes. So far only the CSV filename & metadata have been updated, but we'll also need to make some updates to reflect the switch for margin-of-error (MOE) to high/low confidence intervals for several variables, as described in the Google Doc, once `demographics.shp` has been updated and uploaded to gs.earthmaps.io.


Joshdpaul:
I revised the field descriptions for the CSV output, and also added a new field with info about the source. Descriptive text here was copied verbatim from Micah's suggestions in [this spreadsheet](https://docs.google.com/spreadsheets/d/1GjqBWDhnBOXVeQbiCexvM6WbJNxgTxhY/edit?gid=1512438943#gid=1512438943).

Tested Anchorage in the regular API output (http://localhost:5000/demographics/AK15 , not affected) and the CSV output (http://localhost:5000/demographics/AK15?format=csv , now contains the updated descriptions and new source column).

Joshdpaul:
@cstephen As discussed in the Slack, we implemented the conversion of margin-of-error (MOE) to high/low confidence intervals on the web app side via [this PR](https://github.com/ua-snap/northern-climate-reports/pull/684) instead of on the API side. So this PR just concerns the demographics CSV content.

We can merge these CSV changes now, and potentially revise the `demographics.shp` & upload to `gs.earthmaps.io` in the future.

cstephen:
@Joshdpaul, I did a side-by-side comparison of the spreadsheet they provided vs. the API CSV. The "Data source" column values look good, but I noticed some ambiguity in the spreadsheet they provided. There are effectively two different "new description" columns. Looks like you updated the descriptions based on the first "new description" column, but there's another description column highlighted in green (second to the last column).

Based on the yellow comment at the top of their spreadsheet, the second-to-last description column highlighted in green is their preferred version. This version shortens the description column by removing data source information. Looks like we'll want to use this one. I 100% understand the confusion though!

PR #507, opened 2025-01-07
Replace Physiography with Ecoregions

### LINKED ISSUE #434: Change "Physiography" to Ecoregion
The scope of the term "physiography" ([physical geography](https://en.wikipedia.org/wiki/Physical_geography)) is very broad and encompasses many of the other existing data API service categories (hydrology, permafrost/soils, climate, geology...). 

Since this data category currently only contains the EPA Ecoregions data, consider changing the service category name to "Ecology", "Biogeography", or something similar.

Joshdpaul:
This PR is a simple replacement of "Physiography" with "Ecoregions" everywhere that term occurs in the codebase, excluding the WFS data fetch from GeoServer that still uses "physiography" as the name of the workspace.

Closes #434 

To test, fire up the API and be sure to set environment variables like so:
```
export API_GS_BASE_URL=https://gs.earthmaps.io/geoserver/
export API_RAS_BASE_URL=https://zeus.snap.uaf.edu/rasdaman/
```

Ensure that "physiography" has been changed to "ecoregions" in all cases and that the endpoint works as expected with the new name.


BobTorgerson (approved):
I removed the last of the Physiography references in the repo except for the call to the Geoserver store called "physiography" that needs to remain, unless we changed the name of the store to ecoregions. We could do so if desired, and then we would wdant to change https://github.com/ua-snap/data-api/blob/13e1fce5ef9e0cdb6f3cb1f7735487f40ce04d93/routes/ecoregions.py#L57

Joshdpaul:
Thanks Bob, good catch on the EDS - I didn't even think to look in there for references to physiography!

PR #508, opened 2025-01-08
Adds API endpoint for determining latest month and year for HSIA data.

BobTorgerson:
This PR adds the endpoint /seaice/enddate/ which looks at the ANSI datetime axis of the sea ice coverage to determine the most recent time slice of the data. It returns a JSON object of { "year": <max_year>, "month": <max_month> }

To test, simply go to:

http://localhost:5000/seaice/enddate/

### DIFF
--- a/routes/seaice.py
+++ b/routes/seaice.py
@@ -119,7 +119,5 @@ def seaice_enddate():
     try:
         valid_date = validate_seaice_timestring(latest_date)
         return {"year": valid_date.year, "month": valid_date.month}
-    except ValueError as e:
-        print(e)
-        # Defaults to known good data if the latest date is invalid
-        return {"year": 2024, "month": 6}
+    except Exception as exc:
+        return render_template("500/server_error.html"), 500
--- a/routes/seaice.py
+++ b/routes/seaice.py
@@ -119,5 +119,5 @@ def seaice_enddate():
     try:
         valid_date = validate_seaice_timestring(latest_date)
         return {"year": valid_date.year, "month": valid_date.month}
-    except Exception as exc:
+    except Exception:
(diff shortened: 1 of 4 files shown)


routes/seaice.py
```diff
@@ -92,3 +101,25 @@ def run_point_fetch_all_seaice(lat, lon):
...
+
+    """
+    hsia_encodings = asyncio.run(get_seaice_metadata())
+    latest_date = hsia_encodings["ansi"][-1]
+
+    try:
+        valid_date = validate_seaice_timestring(latest_date)
+        return {"year": valid_date.year, "month": valid_date.month}
+    except ValueError as e:
+        print(e)
```
brucecrevensten, line 123: Remove the print (at least, issue it to the logger).

I feel like this should throw a 503 error, not return a data stub.

routes/seaice.py
```diff
@@ -92,3 +101,25 @@ def run_point_fetch_all_seaice(lat, lon):
...
+        # Defaults to known good data if the latest date is invalid
+        return {"year": 2024, "month": 6}
```
brucecrevensten, line 125: I'm not sure we want to do this: the app kinda lives or dies based on reliable API access.  Perhaps we make a ticket such that if THIS request fails in the client, the client says "Super sorry, I'm ruined" & hides the map/community selectors -- ?

brucecrevensten (commented):
Documentation to add:

 - Document this endpoint in the sea ice doc page
 - Update the sea ice doc page to point to the NSIDC upstream
 - Indicate (probably right before the endpoint documentation) "This dataset is updated periodically throughout the year.  The latest date available can be retreived with this request..."

brucecrevensten (approved):

PR #512, opened 2025-01-13
Fix cmip6 indicators bbox validation

### LINKED ISSUE #493: Fix BBOX validation for CMIP6 monthly and CMIP6 indicator endpoints

charparr:
This PR makes a small modification to the CMIP6 Indicators route such that it fetches the spatial bounds from rasdaman and constructs a bounding box to validate the lat-lon of the request.

This should enable locations to have both their CMIP6 data and CMIP6 indicators queried.

To test this PR, make sure you are pointing at datacubes and then try routes like this:

You should get data here:
http://localhost:5000/indicators/cmip6/point/50.5/-147

But not here:
http://localhost:5000/indicators/cmip6/point/49.5/-147

For the latter, the map isn't great but IMO that is OK for now. Also, the bounding box reported on this page **is accurate** due to the weirdness of the re-grid. However, any requests for latitude 90.1 or something will get caught by this block:

```python
    validation = latlon_is_numeric_and_in_geodetic_range(lat, lon)
    if validation == 400:
        return render_template("400/bad_request.html"), 400
```

Closes #493

### DIFF
--- a/routes/indicators.py
+++ b/routes/indicators.py
@@ -382,7 +382,7 @@ def run_fetch_base_indicators_point_data(lat, lon):
         lon (float): longitude
 
     Returns:
-        JSON-like dict of requested ALFRESCO data
+        JSON-like dict of requested data
 
     Notes:
         example request: http://localhost:5000/indicators/base/point/65.06/-146.16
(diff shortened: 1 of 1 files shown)


cstephen (approved):
Works great! Thanks for the quick fix, @charparr!

PR #534, opened 2025-02-07
Fixes errors with permafrost and beetles coverages near coastal and Canadian communities

### LINKED ISSUE #533: Coastal communities are returning slowly from cache due to permafrost and hydrology
It was noticed that many communities near the coast are slow even when they should be in cache; it looks like this has something to do with permafrost and beetles coverages.

BobTorgerson:
This PR fixes the slowdowns seen by the cache returns for coastal and Canadian communities in the NCR for both the permafrost and beetle API endpoints. 

In both cases, we were returning 500 errors that get translated as 503s from the cache, and these 500 errors were not being cached, thus causing a constant need to retry those endpoints on each request. They both had logic errors that required a small fix to allow for them to return 404 errors which is really what these endpoints are supposed to return for these locations. 

Closes #533

### DIFF
--- a/config.py
+++ b/config.py
@@ -4,8 +4,8 @@ Configuration common to multiple routes.
 
 import os
 
-GS_BASE_URL = os.getenv("API_GS_BASE_URL") or "https://gs.mapventure.org/geoserver/"
-RAS_BASE_URL = os.getenv("API_RAS_BASE_URL") or "https://apollo.snap.uaf.edu/rasdaman/"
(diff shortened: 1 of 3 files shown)


cstephen (approved):
Looks good! I tried a handful of communities and compared the HTTP error codes received from the API between production vs. local Northern Climate Reports pointed at this API branch. I've confirmed that HTTP 500 errors are now showing up as HTTP 404 errors where appropriate for both the beetles and permafrost endpoints.

PR #541, opened 2025-02-14
Replace all instances of <strong> to fix dark mode issues

### LINKED ISSUE #539: Fix dark mode issue on Climate Indicators and Sea Ice Concentration documentation pages
Image

Image

cstephen:
Closes #539.

I did some research and it sounds like the dark mode issue relating to the `<strong>` element is an issue/bug with Bulma itself. But because Bulma provides a `has-text-weight-bold` text-weight typography helper, it's easy to get around this by replacing instances of `<strong>` with `<span class="has-text-weight-bold">`, so that's what this PR does on the following pages:

http://127.0.0.1:5000/indicators/
http://127.0.0.1:5000/seaice/

I also discovered that two of our error templates had the same problem and fixed those as well:

http://127.0.0.1:5000/degree_days/heating/45.0628/-136.1627 (invalid lat/lon error page)
http://127.0.0.1:5000/landfastice/point/50.6123/-139.8869 (outside of coverage error page)

I also went ahead and removed the `<strong>` element from the Elevation documentation since it was being overridden by `<code>` element nested below it and wasn't actually doing anything, and I figure this helps future-proof us from unpredictable behavior:

http://127.0.0.1:5000/elevation/

So, this PR basically just removes all `<strong>` elements from all documentation pages and replaces them where needed.

### DIFF
--- a/templates/422/invalid_latlon.html
+++ b/templates/422/invalid_latlon.html
@@ -11,17 +11,17 @@
 </p>
 <h4 class="mt-4">Western bounding box:</h4>
 <p>
-  <strong>Latitude:</strong> <code>{{ west_bbox[1] }}</code> to <code>{{ west_bbox[3] }}</code>
+  <span class="has-text-weight-bold">Latitude:</span> <code>{{ west_bbox[1] }}</code> to <code>{{ west_bbox[3] }}</code>
 </p>
 <p>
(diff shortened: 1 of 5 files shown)


Joshdpaul:
Tested in Chrome and Safari dark mode and it looks great to me, Craig!

PR #542, opened 2025-02-14
Add CMIP6 reference page

### LINKED ISSUE #444: Review/fix CMIP6 indicators documentation

Joshdpaul:
This PR adds a standalone CMIP6 reference page linked from CMIP6 documentation. 

To test, start the API as usual and review the CMIP6 documentation. Click some of the DOI links to make sure they point to the right dataset, and that the author information is correct.

Closes #444

### DIFF
--- a/routes/cmip6.py
+++ b/routes/cmip6.py
@@ -208,6 +208,11 @@ def cmip6_about():
     return render_template("/documentation/cmip6.html")
 
 
+@routes.route("/cmip6/references")
+def cmip6_references():
+    return render_template("/documentation/cmip6_refs.html")
+
+
 @routes.route("/cmip6/point/<lat>/<lon>")
 @routes.route("/cmip6/point/<lat>/<lon>/<start_year>/<end_year>")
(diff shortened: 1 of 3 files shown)


Joshdpaul:
XREF https://github.com/ua-snap/ardac-explorer/issues/196

cstephen (changes requested):
Having all of this CMIP6 citation info & links in one place is a great addition to the API! 

Most everything looks good to me, but I'm curious about the "E3SM-1-1" and "E3SM-2-0" citation info near the end of the page. These models are not currently included in the CMIP6 monthly or CMIP6 indicators API output, so I'm thinking we should omit these from the documentation until they are available (assuming we are working to include these models).

Joshdpaul:
Good point, Craig. I believe these models were late additions to the ensemble and are important to the InteRFACE project (@brucecrevensten can explain more on that I think) but you are correct that our current CMIP6 coverages do not include data from these models yet. (We have the raw data from the models transferred, but I don't think that data has gone thru the processing pipeline.)

So yeah, referencing the models without having the data available is probably not the best move. I'll comment those out of the HTML until we have the data available.

Issue #536, opened 2025-02-12
Merge "Ecoregions" endpoint into "Physical and Administrative Boundary Polygons"


Joshdpaul:
The Ecoregions endpoint currently just returns the name of an ecoregion that surrounds an input lat/lon. This is not super useful. We could potentially merge this endpoint with the boundarys/areas endpoint, and return the actual GeoJSON polygon for the ecoregion so it can be used in analysis.

PR #543, opened 2025-02-14
Places & Boundary documentation

### LINKED ISSUE #299: Add documentation endpoint for `/places` routes
Let's add a documentation page for the `/places`.  It's a convenient menu for getting what's live in the API in terms of "where can we query?"  ...and provides an easy lookup for questions like "exactly what lat/lng is used for (x)?"

### LINKED ISSUE #300: Review documentation endpoint `/boundary` endpoint
The `/boundary` endpoint looks like it may not be complete.  And it could benefit from some additional discussion about what these exactly are/aren't in terms of some of the ways we've simplified the polygons, etc.  And links to authoritative references, where it makes sense.

### LINKED ISSUE #418: Adjust documentation for boundaries
Right now on [this page](https://earthmaps.io/boundary/) the 2nd section explicitly lists some places but omits HUC10.

I think perhaps the better way to do the documentation for this block -- to reduce things getting missed -- is just to say "use the code found from the section above" with a slew of examples.

### LINKED ISSUE #450: Include source data references for Boundary Data
I thought of this after finding out that the HUC polygons and IDs are a versioned dataset with substantial differences.

There are multiple categories of boundary data that may change over time, and so it's probably best practice to identify which version of the datasets we are serving or at least identify the source of the vector datasets. Sources/versions are not catalogued in the GVV repo, the data API, or in NCR (as far as I can tell).

- HUCs
- AK Native Corp land (there is actually a lot of land that still has yet to be conveyed, and so these boundaries do change!)
- Fire Mgmt Zones

### LINKED ISSUE #500: Add documentation for remaining places
In #499 we got the documentation for where our new Yukon polygons came from.  But we didn't document our other data sources too well.  We need to add where we fetched those polygons from.

Joshdpaul:
This PR establishes 2 separate documentation pages for the `/places` and `/boundary` endpoints. Source data references are included in the documentation for `/places`, and this page is referenced in the source data section of the `/boundary` endpoint. 

This PR closes #450 , closes #500 , closes #418 , closes #299 , closes #300

This PR also addresses (but does not close) #536 and #536 by removing the GeoJSON request example for a HUC12, and removing the "Ecoregions" endpoint from the service categories. Read these 2 issues to see how we can fix these endpoints to make them more consistent with the other boundary data.

### DIFF
--- a/application.py
+++ b/application.py
@@ -16,7 +16,7 @@ app.register_blueprint(routes)
 
 def get_service_categories():
     """
-    This is the new location for the service_categories on the main page.
+    This is the location for the service_categories on the main page.
     This will function will be called on the default route and the list
     will be passed to the index.html template.
     """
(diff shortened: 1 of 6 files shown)


charparr:
@Joshdpaul could you please provide a list of routes that I should test here? I'm a little unsure of what I should be expecting to see. For example, I see that this route looks great

http://127.0.0.1:5000/places/

but based on the changes to `application.py` I feel like I should see a "Geospatial" category on the index page (http://127.0.0.1:5000/) but I'm not seeing one, although I'm not sure that is the behavior I should be testing for.

Joshdpaul:
@charparr My apologies if this was unclear! These are the affected endpoints (just documentation pages, no data-fetching endpoints):

http://127.0.0.1:5000/
http://127.0.0.1:5000/places
http://127.0.0.1:5000/boundary

charparr:
Thanks Josh! The templates look good, and this is a nice addition. I reviewed the documentation to verify the links are valid, and tested the endpoints to get some GeoJSONs back, e.g., http://127.0.0.1:5000/boundary/area/NC8.

charparr (changes requested):
Looking good! Thanks for bearing with me. I left two minor typo / pluralization suggestions.

PR #544, opened 2025-02-14
Use "cmip5" in addition to "base" for indicators

### LINKED ISSUE #515: Change CMIP5 indicators route from `base` to `cmip5`
The indicators route now has both CMIP5- and CMIP6-based data. We should offer these in two separate routes, e.g.:

`/indicators/cmip6/point/61.5/-147`
`/indicators/cmip5/point/61.5/-147`

This requires changing the `base` route to `cmip5`, and will have downstream effects in our applications that pull CMIP5 indicators.

Joshdpaul:
This PR adds an alternate "cmip5" route to the indicators endpoint, and includes this route in the API documentation for consistency with the new "cmip6" indicators route.

The old "base" route is left intact, so that no apps break. 

To test, start the API as usual and review the climate indicators documentation. Test some links using the new "cmip5" option, e.g. http://localhost:5000/indicators/cmip5/area/1903040601

closes #515

### DIFF
--- a/routes/indicators.py
+++ b/routes/indicators.py
@@ -380,7 +380,12 @@ def run_fetch_cmip6_indicators_point_data(lat, lon):
         return render_template("500/server_error.html"), 500
 
 
-@routes.route("/indicators/base/point/<lat>/<lon>")
+@routes.route(
+    "/indicators/base/point/<lat>/<lon>"
+)  # original route, kept for backwards compatibility
+@routes.route(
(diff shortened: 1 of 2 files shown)


cstephen (approved):
This looks good, and I've confirmed that both the old and new versions of the endpoint work as expected. This is a great change, too. The "base" URL path component has been fairly mysterious since we added CMIP6 indicators, and this makes the CMIP5 vs. CMIP6 distinction crystal clear.

PR #545, opened 2025-02-19
Upgrade to 3.11

### LINKED ISSUE #506: Upgrade Python version
Python 3.8 is now EOL: https://devguide.python.org/versions/

The pip environment in the API currently requires 3.8: https://github.com/ua-snap/data-api/blob/7f61bbecb52a6d505d4fd37232d5469d35934e87/Pipfile#L30

Joshdpaul:
This PR upgrades the API to Python 3.11, as the current 3.8 spec is EOL. Closes #506 

In order for this to work, I had to remove the `netcdf4` package, and I unpinned the `requests` version too. After testing all endpoints with non-cached coordinates, everything appears to work as before. I anticipated more problems, but removing that package seemed to have no effect.

To test:

create a new conda environment with Python 3.11 like so (or just use an existing environment with 3.11):
```
conda create -n "api_311" python=3.11
```

Activate that environment, delete any old `Pipfile.lock` file (this step might be unecessary), and reinstall the pip environment:
```
pipenv install
```

Then run the API as usual. If successful , all endpoints should work as before.  Is there anything I missed here? Any additional testing we could do?


cstephen (approved):
I installed this branch's pipenv environment like this:

```
pipenv --rm
pipenv install
```

And confirmed that I was getting the right version of Python:

```
$ pipenv run python -V
Python 3.11.9
```

I ran into problems running `pipenv` using a Python 3.11.9 conda environment for some reason, but I think as long as the `pipenv` command above is reporting Python 3.11.9, everything is working as expected.

I then ran the Data API from this branch and pointed local instances of the following webapps against it:

- Northern Climate Reports
- Arctic-EDS
- Historical Sea Ice Atlas
- ARDAC Explorer

And everything is working as it should! This covers a ton of test cases all at once, so I think all functionality survived the Python upgrade (and removal of the `netcdf4` package too).

I searched through the code and the only references I can see to the `netcdf4` package are in the `Pipfile`, `Pipfile.lock`, and `requirements.txt` files. So, I think this was probably a package we stopped using and forgot to remove from the Pipfile. That would explain why all functionality seems to be intact without it.

So, everything looks good with this branch! But we may need to also generate a new `requirements.txt` file with the following command to make sure our Elastic Beanstalk instance uses package versions compatible with Python 3.11.9:

```
pip freeze > requirements.txt
```

I'll defer to @BobTorgerson on this since he has more experience with this kind of thing.

PR #548, opened 2025-02-27
Incorporate ecoregions + HUC12's in new `all_areas` layer into API

Joshdpaul:
This PR revises the API to pull ecoregion and HUC12 polygons from the new `all_areas` layer (see [here](https://github.com/ua-snap/geospatial-vector-veracity/pull/148)). This branch uses some placeholder code to direct requests to this new layer (`playground:all_areas`) on the development server. Eventually this layer will replace the `all_boundaries:all_areas` layer on the production server, and we can drop the placeholder code.

Previously, ecoregion polygons were unavailable for area queries. By adding ecoregion polygons to the `all_areas` layer, these ecoregions now become available to summarize datasets.

Previously, HUC12 polygons were queried from a completely separate layer, and so requests needed to be parsed to recognize if a request was for a HUC12 or not. The revisions in this branch simply direct all area queries to the standard `all_areas` layer, and so that conditional code could removed. There is still a function specifically devoted to fetching HUC12 polygons for the ALFRESCO point data summary, but the CQL query used in this function now specifies the new `all_areas` layer, plus an additional conditional statement to request only the HUC12 `area_type`. 

closes https://github.com/ua-snap/data-api/issues/536 and closes https://github.com/ua-snap/geospatial-vector-veracity/issues/146

To test:

Start the API as usual, but point GeoServer queries to the development server like so:
```export API_GS_BASE_URL=https://gs.mapventure.org/geoserver/```

- Check the `places` documentation at http://127.0.0.1:5000/places/ to make sure you see both ecoregions and HUC12s described in the service categories and in the source references.
- Test the http://127.0.0.1:5000/places/all endpoint to make sure it includes HUC12s and ecoregions, and that the CSV option does the same http://127.0.0.1:5000/places/all?format=csv
- Test the endpoints (http://127.0.0.1:5000/places/hucs and http://127.0.0.1:5000/places/ecoregions) to make sure they work as expected, and that the CSV options do the same (http://127.0.0.1:5000/places/hucs?format=csv , http://127.0.0.1:5000/places/ecoregions?format=csv)
- Check the `boundary` documentation at http://127.0.0.1:5000/boundary/ to make sure you see both ecoregions and HUC12s described in the service categories, and make sure the endpoints return GeoJSON for valid ecoregion and HUC12 IDs.
- Check the text changes in the `alfresco` documentation http://127.0.0.1:5000/alfresco/ - these are minor but I think they improve consistency with the HUC data source described in the `places` endpoint
- Make sure the `alfresco` point queries still return data, and test HUC12s and ecoregions to make sure they return data in the area query (e.g. http://127.0.0.1:5000/alfresco/flammability/area/AKECO1 ), and that the CSV option does the same (e.g., http://127.0.0.1:5000/alfresco/flammability/area/AKECO1?format=csv )
- Test any other area-based query to make sure it returns data for HUC12 and ecoregion IDs. Note that some datasets may not return any data for the HUC12 polygons since they can be quite small. In these cases you should see NaN returns, and not errors.

### CODE REVIEW

**Review by charparr — CHANGES_REQUESTED** [review]
Great PR! Easy to follow the description, and the links all work as I'd expect. I'm seeing the HUC-12s and Ecoregions where they should be, and it was a great idea to consolidate how we treat these polygons. Nice one. 

I'd support an additional commit here that uses an environment variable for the GeoServer target store - most of my comments are along those lines. Curious to know what @cstephen thinks regarding that approach.

**Comment by Joshdpaul** [comment]
@charparr Thanks for the review! Great suggestions, you started the conversation to get a "real" development server going, which is awesome. 

I believe I've addressed all your comments with [dabdf71](https://github.com/ua-snap/data-api/pull/548/commits/dabdf710611a2334daab6c80ff8ef7c0baa287d4).

To re-test:

Do the same as the original instructions, but point GeoServer queries to the new development server like so:
```
export API_GS_BASE_URL=https://gs-dev.earthmaps.io/geoserver/
```

charparr (approved):
OK I've re-reviewed using the development GeoServer that Bob set up yesterday:

API_GS_BASE_URL=https://gs-dev.earthmaps.io/geoserver/

And everything is working great and I can confirm that the commented-out snippets are no longer present. I think it is ready to go!

cstephen (approved):
I tested this with a fine-toothed comb, gave the documentation a careful read, tested documentation links, tried all of the test cases in the PR description and variants with other Ecoregion and HUC-12 area codes + other area summary endpoints. Everything looks & behaves perfectly! 💯 

Sounds like the discussion about setting the GeoServer store as an environment variable is no longer relevant now that we're pointing this at a proper GeoServer development server. Seems like a pretty good idea, but not needed for this PR, so I recommend skipping this for now and merging this PR as-is.

PR #553, opened 2025-03-10
Perform local lat/lon check against GeoTIFF coverage masks to check for nodata before fetching from Rasdaman

### LINKED ISSUE #532: Use an example GeoTIFF from each dataset to do a local check for nodata pixels on the API side before sending a request to Rasdaman
The idea here is prevent wasteful requests to Rasdaman that we know will not return anything meaningful.

cstephen:
Closes #532.

This PR adds a GeoTIFF mask file for nearly every Rasdaman coverage used by the Data API. The API uses these files to do a local lat/lon check against the GeoTIFF to verify that data will be available before sending a data fetch to Rasdaman. This has a couple big benefits:

- Lat/lon coordinates within a coverage's BBOX, but with no data, now tend to return in a fraction of a second (instead of up to 60+ seconds in some cases). [See benchmarks here](https://docs.google.com/spreadsheets/d/1hBWgD7WOB8GF91cRtUIOWBhxuNlqfKYlVQ_KqNjPKz4/edit?gid=0#gid=0).
- Filters out wasteful nodata requests before sending them to Rasdaman, putting less load on Rasdaman

This PR only checks GeoTIFFs for single lat/lon points, not areas, because deciding what to do about an area with partial nodata is not obvious, and also performing pixel intersection operations with a local GeoTIFF could add (I think) a non-trivial amount of extra work to each area request. But we can revisit this later if needed.

I've omitted the ALFRESCO endpoints from the local GeoTIFF check for the reason mentioned above. Since all ALFRESCO point queries get turned into HUC-12 area queries, it's not clear how to handle this or whether it's worth it.

I've also omitted local GeoTIFF checks for the CMIP6 monthly and CMIP6 indicators coverages since Rasdaman deals with the antimeridian nodata "ray of hope" issues better than a local GeoTIFF check, often returning valid data, whereas a local GeoTIFF check returns nodata. Plus, the CMIP6 data footprint is basically just a rectangular BBOX that is better handled via BBOX lat/lon validation.

I've also had to tweak every impacted endpoint to recognize the new 404 error code that will result from finding nodata in a local GeoTIFF.

Testing this PR is tricky because there are so many endpoints to test, and deciding which lat/lon coordinate to check depends largely on the endpoint/coverage being tested. To make this easier, I've whipped up the following throwaway Python script that tests all affected endpoints with random lat/lon coordinates within a BBOX (loosely) around Alaska. This BBOX covers a lot of both land & ocean, so it's a good way to test land- and ocean-based datasets at the same time.

```python
#!/usr/bin/env python
import requests
import time
import numpy as np
import sys

domain = "http://127.0.0.1:5000"
prod_domain = "https://earthmaps.io"
requests_per_endpoint = int(sys.argv[1])
bbox = [-170, 52, -134, 72]

endpoints = [
    "/beetles/point",
    "/degree_days/heating",
    "/degree_days/below_zero",
    "/degree_days/freezing_index",
    "/degree_days/thawing_index",
    "/hydrology/point",
    "/landfastice/point",
    "/indicators/base/point",
    "/permafrost/point/gipl",
    "/precipitation",
    "/precipitation/frequency/point",
    "/seaice/point",
    "/snow/snowfallequivalent",
    "/tas2km/point",
    "/temperature/jan",
    "/temperature/july",
    "/temperature",
    "/taspr/point",
    "/wet_days_per_year/hp/point",
]


for endpoint in endpoints:
    for i in range(requests_per_endpoint):
        lat = round(np.random.uniform(bbox[1], bbox[3]), 2)
        lon = round(np.random.uniform(bbox[0], bbox[2]), 2)
        latlon_subpath = f"/{lat}/{lon}"

        dev_url = domain + endpoint + latlon_subpath
        start_time = time.time()
        try:
            response = requests.get(dev_url)
            dev_status = response.status_code
        except:
            dev_status = "Failure"
        end_time = time.time()
        dev_time = round(end_time - start_time, 2)

        prod_url = prod_domain + endpoint + latlon_subpath
        start_time = time.time()
        try:
            response = requests.get(prod_url)
            prod_status = response.status_code
        except:
            prod_status = "Failure"
        end_time = time.time()
        prod_time = round(end_time - start_time, 2)

        if dev_status != prod_status:
            print(f"{endpoint}{latlon_subpath}")
            print(f"dev: HTTP {dev_status}, {dev_time} seconds")
            print(f"prod: HTTP {prod_status}, {prod_time} seconds")
            print("----------")

        time.sleep(3)
```

```bash
Usage: python benchmark.py <number of requests per endpoint>
Example: python benchmark.py 5
```

The script runs silently until it encounters any endpoint where the HTTP status code differs between the local (development) API and our current production data, which may be worth investigating further. While running this script, I've found a few interesting discrepancies between local vs. production API:

- The production API will sometimes return a large JSON response full of nulls (nodata). This might be due to missing a `nullify_and_prune` call or two, but I think `null`s are also getting wrapped in quotation marks when they weren't before (i.e., before vs. after a recent Rasdaman upgrade) and aren't detected as proper nodata for this reason. Example: [https://earthmaps.io/tas2km/point/57.59/-151.81](https://earthmaps.io/tas2km/point/57.59/-151.81) The local GeoTIFF check solves this in a different way, returning the HTTP 404 code/page as expected.
- Sometimes a production endpoint will return an HTTP 503 error if the WCS Rasdaman fetch times out, if nodata responses aren't properly handled in the API code, etc. The local GeoTIFF check improves some of these cases by returning an HTTP 404 code/page for nodata requests.

But in general, any lat/lon endpoint that returns a HTTP 200 (data returned) on the production server, and returns a non-HTTP-200 code for the same endpoint on the development server, is worth paying attention to. If the production server response is full of nothing but nulls, this is an improvement. If the production server response is full of actual data, this is a problem I'll need to fix.


charparr:
I'm loving the preliminary check of the mask GeoTIFF to ascertain whether or not data exists for the point. This is nifty.
It certainly alleviates the frustration of having a long-running request yield nothing - that is a bummer for us and other users, and seems to have a negative impact on the Rasdaman server.

This will be a longer review, so I"m going to chunk it up across several comments based on the status code pairs between dev and prod. All results are from running the supplied benchmark script, 5 requests per endpoint. Output of my script is attached.
[parr_benchmark_03132025.txt](https://github.com/user-attachments/files/19232281/parr_benchmark_03132025.txt)

charparr:
### Dev is 422, prod is 503
We'd generally assume that any change from 503 is good.
```
/degree_days/thawing_index/71.75/-145.25
dev: HTTP 422, 0.01 seconds
prod: HTTP 503, 0.27 seconds

/snow/snowfallequivalent/71.4/-143.23
dev: HTTP 422, 0.01 seconds
prod: HTTP 503, 0.29 seconds

/wet_days_per_year/hp/point/71.92/-159.53
dev: HTTP 422, 0.01 seconds
prod: HTTP 503, 0.29 seconds

/wet_days_per_year/hp/point/71.99/-137.45
dev: HTTP 422, 0.01 seconds
prod: HTTP 503, 0.27 seconds
```

In all these cases the points are just a little too far North and not within the `EAST_BBOX` or `WEST_BBOX`. In dev, these endpoints all now render the same template with the bounding box map and list of bbox vertices. This is definitely an improvement in the status code returned, and also in speed. My suggestion though is that each of these endpoints should inject their own bbox into the template - we have this behavior in a few of our other endpoints. I'm not really sure why these were 503'ing in the first place, but if you look at the diff for `wet_days_per_year` for example you see that there was never really a good catch for the situation in which the validation function returned a 404.

**Conclusion: based on the endpoints in my test, every scenario in which dev returns 422, but prod returns 503, the 422 is the desired behavior.**

charparr:
#### Dev is 500, prod is 200

OK, this is also a scenario where we need to check carefully. We're tossing out 500s in dev, but prod says OK. What is happening?

```
/precipitation/68.3/-163.49
dev: HTTP 500, 74.2 seconds
prod: HTTP 200, 14.28 seconds

/precipitation/65.35/-144.8
dev: HTTP 500, 3.44 seconds
prod: HTTP 200, 14.77 seconds

/precipitation/58.67/-155.29
dev: HTTP 500, 74.39 seconds
prod: HTTP 200, 14.97 seconds

/precipitation/frequency/point/65.36/-146.08
dev: HTTP 500, 38.61 seconds
prod: HTTP 200, 31.0 seconds

/precipitation/frequency/point/67.81/-143.68
dev: HTTP 500, 23.32 seconds
prod: HTTP 200, 30.59 seconds

/precipitation/frequency/point/69.73/-163.04
dev: HTTP 500, 1.08 seconds
prod: HTTP 200, 26.44 seconds

/snow/snowfallequivalent/64.26/-152.63
dev: HTTP 500, 92.07 seconds
prod: HTTP 200, 2.01 seconds

/temperature/64.31/-158.35
dev: HTTP 500, 74.44 seconds
prod: HTTP 200, 14.56 seconds

/temperature/66.0/-147.88
dev: HTTP 500, 258.88 seconds
prod: HTTP 200, 14.97 seconds
```
*We can ignore these! I hit these endpoints again in dev and it yielded identical data responses to prod. False positives here that are just an artifact of running the API locally. For further benchmarking, add a retry or sleep when the request is made to the local API.*
```
/tas2km/point/59.63/-160.61
dev: HTTP 500, 10.21 seconds
prod: HTTP 200, 19.4 seconds

/tas2km/point/62.6/-134.09
dev: HTTP 500, 38.09 seconds
prod: HTTP 200, 276.44 seconds

/tas2km/point/65.14/-157.07
dev: HTTP 500, 34.58 seconds
prod: HTTP 200, 263.3 seconds
```
*These are all terrestrial locations with valid data responses in prod. However, after a few re-tries I was able to replicate the prod response in dev, so this is similar to the above. However, just a note that these data responses yield strings-of-floats like this, which seems weird:*
```
{
  "historical": {
    "CRU-TS": {
      "historical": {
        "April": {
          "1901": {
            "tasmax": "-0.6",
            "tasmean": "-6.8",
            "tasmin": "-12.9"
          },
```
**Conclusion: based on my testing the scenario in which dev is 500 but prod is 200 is a red herring caused by the local API instance taking a little longer.**

charparr (approved):
This PR is a substantial improvement and in every case in which the dev and prod API differed in their status codes, I prefer the dev response. Based on my testing, this is good to go!

However, this PR does shine a light on some issues with validation. For starters, we could probably consolidate how we validate requests across different endpoints. Theoretically, every `/point/` endpoint should have the same validation pattern. I don't think that should let us get distracted from what we're doing in this PR though.

Also, for anyone else testing this PR, you may have trouble replicating some of the cases where the production API was yielding 503. There is some stuff in-flight with the Varnish Cache (see dicussion in #api in Slack) where we will miss our fancy 400 or 422 error messages in the production API for a bit.

XREF #555

Joshdpaul:
I don't have much to add here, @charparr 's testing is thorough and notes all of the improved validation behaviors. The results of my testing did not bring anything else to light. I also approve! This was a great idea @cstephen and definitely a successful effort with regard to speeding up the validation, and as a bonus we caught an issue with the Varnish cache HTTP response status codes. Wins all around 🏆 !

I think the 404 in development, 200 in production situation where we are getting all `null` or `nan` in the production JSON return might be traced back to the [`postprocessing.nullify_and_prune(...)`](https://github.com/ua-snap/data-api/blob/ff6d09e931d13912b07c6c8b2229ac4baa7f0519/postprocessing.py#L176) and/or the [`postprocessing.prune_nulls_with_max_intensity(...)`](https://github.com/ua-snap/data-api/blob/ff6d09e931d13912b07c6c8b2229ac4baa7f0519/postprocessing.py#L141) functions. I have noticed in other dev work that these functions do not always work as intended, and depending on the dictionary structure can sometimes leave the structure intact while being full of no data values. 

On the topic of consolidating / streamlining our validation process, both @charparr and @cstephen have checked out some of my tinkering in [this repo](https://github.com/ua-snap/jp-fastapi). The `pydantic` validation used in there has huge potential, especially if we combine it with a strong(er) Rasdaman metadata.

Joshdpaul (approved):

PR #567, opened 2025-04-30
test cmip6_monthly_e3sm coverage

Joshdpaul:
This PR makes the changes necessary to fetch data from the `cmip6_monthly` coverage on Apollo, which now has E3SM model data included. 

This coverage differs from the production `cmip6_monthly` on Zeus in that it has an encoding dictionary with integer keys rather than string keys. Previously, we had to convert the keys to strings in order to fetch values from the encodings dictionary, but this branch removes that temporary patch. 

For this reason, even though the `validate_request.get_coverage_encodings()` function has been changed to find the encodings dictionary in both production and development coverage metadata (which ensures the app will start no matter which Rasdaman server its pointed at), the actual data fetching from production `cmip6_monthly` coverage will fail if this is merged into `main`. **So we should wait until the production `cmip6_monthly` coverage is replaced with the dev coverage before merging this branch into `main`!**

To test:

Basically we just want to be sure the API will fetch data as usual from the new CMIP6 monthly coverage that contains the E3SM models.

Point the API at Apollo and fetch some data from the CMIP6 endpoint. Search the JSON return for "E3SM" to make sure the data shows up. Try running the same queries from `earthmaps.io/cmip6` and spot checking to make sure the other models are returning the same data. **Note that `pr` and `prsn` variables will return slightly different values between production and development here!** @Kyle and I discussed this and we think the interpolation parameters in the regridding pipeline have changed since the production ingest, and that is the cause of the discrepancy. No need for alarm here, we are going to commit to the newest interpolation method.

Note that there is a small difference in the encodings dictionary between production and dev: the production coverage will return `swe` instead of `snw` as a variable (`swe` data is not actually a CMIP6 variable id - the variable id is technically `snw` which is mathematically equivalent to `swe` .... we may have to rectify this when pushing the `cmip6_monthly` into production).

### DIFF
--- a/routes/cmip6.py
+++ b/routes/cmip6.py
@@ -18,7 +18,7 @@ from . import routes
 
 cmip6_api = Blueprint("cmip6_api", __name__)
 
-cmip6_monthly_coverage_id = "cmip6_monthly"
+cmip6_monthly_coverage_id = "cmip6_monthly_e3sm"
 
 
 async def get_cmip6_metadata():
@@ -99,27 +99,15 @@ def package_cmip6_monthly_data(
         if var_id != None:
             varname = var_id
         else:
-
-            # TODO: fix cryo coverage so we can delete this line below
-            # temporary fix for "string key" issue
-            var_coord = str(var_coord)
             varname = dim_encodings["varname"][var_coord]
-
+        
         for mi, model_li in enumerate(var_li):
-
-            # TODO: fix cryo coverage so we can delete this line below
-            # temporary fix for "string key" issue
-            mi = str(mi)
-
             model = dim_encodings["model"][mi]
+
             if model not in di:
                 di[model] = dict()
+
             for si, scenario_li in enumerate(model_li):
-
-                # TODO: fix cryo coverage so we can delete this line below
-                # temporary fix for "string key" issue
-                si = str(si)
-
                 scenario = dim_encodings["scenario"][si]
                 if scenario not in di[model]:
                     di[model][scenario] = dict()
@@ -175,6 +163,7 @@ def package_cmip6_monthly_data(
     # all 0 values are replaced with -9999 and will be pruned from the response.
     # If the scenario is not historical, and the year is less than 2015,
     # all 0 values are replaced with -9999 and will be pruned from the response.
+                    
 
     for model, scenarios in di.items():
         for scenario, months in scenarios.items():
@@ -295,6 +284,7 @@ def run_fetch_cmip6_monthly_point_data(lat, lon, start_year=None, end_year=None)
                 new_results = package_cmip6_monthly_data(
                     point_data_list, var_id, start_year, end_year
                 )
+
                 for model, scenarios in new_results.items():
                     results.setdefault(model, {})
                     for scenario, months in scenarios.items():
--- a/routes/cmip6.py
+++ b/routes/cmip6.py
@@ -18,7 +18,7 @@ from . import routes
 
 cmip6_api = Blueprint("cmip6_api", __name__)
 
-cmip6_monthly_coverage_id = "cmip6_monthly_e3sm"
+cmip6_monthly_coverage_id = "cmip6_monthly"
 
 
 async def get_cmip6_metadata():
@@ -100,7 +100,7 @@ def package_cmip6_monthly_data(
             varname = var_id
         else:
             varname = dim_encodings["varname"][var_coord]
-        
+
         for mi, model_li in enumerate(var_li):
             model = dim_encodings["model"][mi]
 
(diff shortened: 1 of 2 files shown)


Joshdpaul:
XREF to ARDAC PR -->> https://github.com/ua-snap/ardac-explorer/pull/249

cstephen (changes requested):
I've been testing this against the corresponding ARDAC Explorer branch, and mostly everything is looking good! I haven't finished the ARDAC Explorer PR review yet, but found a few (mostly unrelated) issues that the new models outputs helped to reveal.

Regarding this Data API PR, though, I did find one interesting issue. The JSON output for the E3SM-1-1/ssp245 starts at `2020-01`, whereas all other CMIP6 models/scenarios start at `2015-01`. Even E3SM-1-1/ssp585 starts at `2015-01`, so I'm not sure why the `ssp245` scenario in particular has this issue. This causes problems with chart population in ARDAC Explorer.

If this is due to the source data itself, there are things we can do on the ARDAC Explorer side to smooth over this difference, but I'm hoping there's another explanation. Not sure if this is due to CMIP6 data preprocessing, overly aggressive nodata pruning on the Data API side, or what.

Joshdpaul:
Good catch, Craig! Here is what I found:

Not only does the `ssp245` data start in 2020, it only goes to 2024! Not sure why, but this model is very limited in the time domain. I confirmed this on ESGF metagrid (the data source) and sure enough, we there are only 5 years available. Its crazy that we got this far through the process without noticing! I'm going to make an issue in the CMIP6 repo that checking for year range should be a more explicit part of our initial data audit process (could be useful for CMIP7).

Our holdings:

```
(base) [jdpaul3@chinook04 E3SM-1-1]$ tree ssp245/Amon
ssp245/Amon
├── clt
│   ├── clt_Amon_E3SM-1-1_ssp245_regrid_202001-202012.nc
│   ├── clt_Amon_E3SM-1-1_ssp245_regrid_202101-202112.nc
│   ├── clt_Amon_E3SM-1-1_ssp245_regrid_202201-202212.nc
│   ├── clt_Amon_E3SM-1-1_ssp245_regrid_202301-202312.nc
│   └── clt_Amon_E3SM-1-1_ssp245_regrid_202401-202412.nc
├── pr
│   ├── pr_Amon_E3SM-1-1_ssp245_regrid_202001-202012.nc
│   ├── pr_Amon_E3SM-1-1_ssp245_regrid_202101-202112.nc
│   ├── pr_Amon_E3SM-1-1_ssp245_regrid_202201-202212.nc
│   ├── pr_Amon_E3SM-1-1_ssp245_regrid_202301-202312.nc
│   └── pr_Amon_E3SM-1-1_ssp245_regrid_202401-202412.nc
├── rlds
│   ├── rlds_Amon_E3SM-1-1_ssp245_regrid_202001-202012.nc
│   ├── rlds_Amon_E3SM-1-1_ssp245_regrid_202101-202112.nc
│   ├── rlds_Amon_E3SM-1-1_ssp245_regrid_202201-202212.nc
│   ├── rlds_Amon_E3SM-1-1_ssp245_regrid_202301-202312.nc
│   └── rlds_Amon_E3SM-1-1_ssp245_regrid_202401-202412.nc
├── rsds
│   ├── rsds_Amon_E3SM-1-1_ssp245_regrid_202001-202012.nc
│   ├── rsds_Amon_E3SM-1-1_ssp245_regrid_202101-202112.nc
│   ├── rsds_Amon_E3SM-1-1_ssp245_regrid_202201-202212.nc
│   ├── rsds_Amon_E3SM-1-1_ssp245_regrid_202301-202312.nc
│   └── rsds_Amon_E3SM-1-1_ssp245_regrid_202401-202412.nc
├── sfcWind
│   ├── sfcWind_Amon_E3SM-1-1_ssp245_regrid_202001-202012.nc
│   ├── sfcWind_Amon_E3SM-1-1_ssp245_regrid_202101-202112.nc
│   ├── sfcWind_Amon_E3SM-1-1_ssp245_regrid_202201-202212.nc
│   ├── sfcWind_Amon_E3SM-1-1_ssp245_regrid_202301-202312.nc
│   └── sfcWind_Amon_E3SM-1-1_ssp245_regrid_202401-202412.nc
└── tas
    ├── tas_Amon_E3SM-1-1_ssp245_regrid_202001-202012.nc
    ├── tas_Amon_E3SM-1-1_ssp245_regrid_202101-202112.nc
    ├── tas_Amon_E3SM-1-1_ssp245_regrid_202201-202212.nc
    ├── tas_Amon_E3SM-1-1_ssp245_regrid_202301-202312.nc
    └── tas_Amon_E3SM-1-1_ssp245_regrid_202401-202412.nc
```

ESGF metagrid holdings:

(This JSON is hard to parse, but its the exported search results for this model, `ssp245` scenario, and precipitation variable. If you search for `"datetime_end"` in the JSON, you should see all of the end dates for the files are 2024.)

[search.json](https://github.com/user-attachments/files/20065717/search.json)

cstephen (approved):

PR #568, opened 2025-05-02
Add new Temperature Anomalies endpoint for use with ARDAC climate stripes

cstephen:
This PR adds a new Temperature Anomalies endpoint to the Data API, which is used to populate the new (soon-to-launch) historical + projected climate stripes widget in ARDAC Explorer.

See https://github.com/ua-snap/climate-stripes-preprocessing/pull/1 if you are curious how the dataset was produced.

To view/test this PR, run the Data API while pointed at Apollo:

```
export API_RAS_BASE_URL=https://apollo.snap.uaf.edu/rasdaman/
pipenv run flask run
```

The new endpoint is linked as "Temperature Anomalies" from the front page of the Data API. Please take a look at the following:

- Read the web documentation to make sure it's clear & makes sense.
- Try accessing the JSON data for a few lat/lon points. For example: http://127.0.0.1:5000/temperature_anomalies/point/63.73/-166.32
- View an exported CSV. For example: http://127.0.0.1:5000/temperature_anomalies/point/63.73/-166.32?format=csv


validate_request.py
```diff
@@ -357,7 +357,7 @@ def validate_latlon_in_bboxes(lat, lon, bboxes, coverages=[]):
     for bbox in bboxes:
         valid_lat = bbox[1] <= lat <= bbox[3]
         valid_lon = bbox[0] <= lon <= bbox[2]
-        if valid_lat or valid_lon:
+        if valid_lat and valid_lon:
```
cstephen, line 360: Btw, this is a bug I found/fixed while working on this PR. I'm pretty sure I am responsible for this bug and failed to fix it the same time these other two lines were fixed:

https://github.com/ua-snap/data-api/blob/5b4d056dc20eeb5e5c536510340f2f77c4acb75f/validate_request.py#L99

https://github.com/ua-snap/data-api/blob/5b4d056dc20eeb5e5c536510340f2f77c4acb75f/validate_request.py#L130

Obviously, both the provided latitude and longitude need to be valid to be inside the coverage's BBOX! 🤦 This fix was necessary to get the API to show me the coverage extent error page using the `invalid_latlon_outside_coverage.html` template when an invalid lat/lon is provided.

charparr: Well, I looked at the git blame for that and it was on a PR that I reviewed! So, right there with you 😀 
Good catch here.

routes/temperature_anomalies.py
```diff
@@ -0,0 +1,171 @@
+import asyncio
+from math import floor
```
charparr, line 2:
```suggestion

```
I think this import is not used

routes/temperature_anomalies.py
```diff
@@ -0,0 +1,171 @@
...
+    Package the temperature anomaly values into human-readable JSON format
+
+    Args:
+        point_data_list (list): nested list of data from Rasdaman WCS query
+
+    Returns:
+        di (dict): dictionary mirroring structure of nested list with keys derived from anomaly_dim_encodings global variable
+    """
+    di = dict()
+    years = list(range(1850, 2101))
```
charparr, line 48: Observation: the year min and max could be parsed from the `anomaly_metadata` object, but I kinda like them being hard-coded for readability and because this coverage won't change in that regard.

routes/temperature_anomalies.py
```diff
@@ -0,0 +1,171 @@
...
+    construct_latlon_bbox_from_coverage_bounds,
+    validate_latlon_in_bboxes,
+)
+from generate_urls import generate_wcs_query_url
+from generate_requests import generate_wcs_getcov_str
+from fetch_data import fetch_data, describe_via_wcps
+from postprocessing import merge_dicts, postprocess, prune_nulls_with_max_intensity
+from . import routes
+
+seaice_api = Blueprint("temperature_anomalies_api", __name__)
```
charparr, line 25: Guessing this route was seeded with the sea ice route!

```suggestion
temperature_anomaly_api = Blueprint("temperature_anomalies_api", __name__)
```

or, since this doesn't really get called anywhere we could do something like 

```suggestion
_temperature_anomaly_api = Blueprint("temperature_anomalies_api", __name__)
```

charparr (changes requested):
Very well constructed endpoint (and source dataset) here! Super tidy packaging functions. I tested a bunch of routes and enjoyed the data returns. CSVs look good as well. I was able to error out as I'd expect and get the correct template pages.

See my minor suggestion for where the Blueprint was initialized to an object called `seaice_api` for my only requested change!

charparr (approved):

Joshdpaul:
Great work, this endpoint works as expected and is a nice addition to our API!

Only 1 comment:

- Consider specifying somewhere in the documentation that "temperature" in this endpoint means "air temperature". Seeing as this is a pan-Arctic dataset that does return values over ocean pixels (e.g. http://127.0.0.1:5000/temperature_anomalies/point/56/-145 ) we might want to be clear that we are not talking about water temperature.

Joshdpaul (approved):

cstephen:
> * Consider specifying somewhere in the documentation that "temperature" in this endpoint means "air temperature". Seeing as this is a pan-Arctic dataset that does return values over ocean pixels (e.g. http://127.0.0.1:5000/temperature_anomalies/point/56/-145 ) we might want to be clear that we are not talking about water temperature.

Good call on this, @Joshdpaul. I've added the phrase "surface air temperature" in several places throughout the documentation after double checking that this is the right terminology to use, here: https://berkeley-earth-temperature-hr.s3.amazonaws.com/Gridded/High-Res_Gridded_README.txt

PR #615, opened 2025-08-12
Add Endpoints for WRF Dynamically Downscaled ERA5 4km Data

charparr:
This PR introduces a new ERA5-WRF route `/era5wrf/` with the following features:
- Point endpoint to return daily time-series for selected variables at a lat/lon.
- Area endpoint to return daily zonal-mean time-series over a polygon.
- Both point and area endpoints support variable selection and CSV output.

Here are the variables available:
```
t2_max: Daily max temperature
t2_mean: Daily mean temperature
t2_min: Daily minimum temperature
rainnc_sum: Daily total precip
rh2_max: Daily max 2 m relative humidity
rh2_mean: Daily mean 2 m relative humidity
rh2_min: Daily min 2 m relative humidity
wdir10_mean: Daily average 10 m wind direction
wspd10_max: Daily max 10 m wind speed
wspd10_mean: Daily mean 10 m wind speed
seaice_max: Daily max sea ice concentration
```

**To test this PR:**

- Point your API instance at Zeus
- Navigate to the documentation page and read through it. Does the content reads clearly?
- Try a bunch of the example routes listed in the docs, and try a few of your own. You should see JSONs keyed by `YYYY-MM-DD` at the top level, with values by keyed by variable name. All the example links should work. Try to verify that:
  - When `format=csv` is present, you get a well formatted CSV with rows corresponding to dates and columns for selected variables. 
  - Variable filtering works as expected.

- Try some expected failure routes:
  - Invalid variable: `/era5wrf/point/65/-146?vars=foofoobarbarfooooo` should yield a 400.  

Some other notes:
- We're rounding to 1 decimal for all variables.
- Each variable exists on Zeus as a discrete coverage and so when multiple variables are requested, several coverages are pinged in parallel. Other coverages package all the variables in a single cube, but that isn't really viable here given the data volume. While WCPS does support operations across coverages (e.g., you could subtract t2min from t2max for dtr) and it could be used to create the requests for this endpoint, I elected to go for a more "standard" async data fetch for multiple variables in parallel because it enables more use of existing code and because this way if one variable fails, others might still succeed and if nothing else it should make it easier to identify which variable/coverage has issues.
- The `validate_request.generate_time_index_from_coverage_metadata()` function was originally implemented in the landfast sea ice route, but I've ripped it out of there because it is useful for anything that pings a coverage where the time info is on an axis of `"type": "ansidate"`. It now lives in the shared `validate_request.py` module. I'm confident that the functionality of the landfast sea ice route is intact, but you can test some of those queries too if you like.

### DIFF
--- a/templates/documentation/era5wrf.html
+++ b/templates/documentation/era5wrf.html
@@ -4,7 +4,7 @@
 <p>
   This endpoint provides access to dynamically downscaled (using the Weather
   Research and Forecasting (WRF) model) ERA5 reanalysis climate data. The
-  dataset spans 1960-2023 with daily temporal resolution and 4 km spatial
+  dataset spans 1960&ndash;2023 with daily temporal resolution and 4 km spatial
   resolution. Available data include temperature, humidity, precipitation, wind,
   and sea ice variables. To produce this dataset, ERA5 was used to initialize
   the WRF model version 4.3.3 over the central Alaska landmass and immediately


routes/era5wrf.py
```diff
@@ -0,0 +1,559 @@
...
+
+        # Extract coordinate list and get bounds
+        time_coordinates = time_axis["coordinate"]
+
+        # Convert to datetime objects
+        min_date = pd.to_datetime(time_coordinates[0])
+        max_date = pd.to_datetime(time_coordinates[-1])
+
+        # Remove timezone info to make naive datetimes
+        min_date_naive = (
```
brucecrevensten, line 116: Possible concern here, what's the time zone for the incoming ERA5 data?  Why would we shift these to be naive?

routes/era5wrf.py
```diff
@@ -0,0 +1,559 @@
...
+            min_date.tz_localize(None) if min_date.tz is not None else min_date
+        )
+        max_date_naive = (
+            max_date.tz_localize(None) if max_date.tz is not None else max_date
+        )
+
+        return min_date_naive.to_pydatetime(), max_date_naive.to_pydatetime()
+
+    except (KeyError, StopIteration, IndexError):
+        raise ValueError(
```
brucecrevensten, line 126: Can you trigger this error intentionally?  Who eats it, where is it emitted?  In logs?

routes/era5wrf.py
```diff
@@ -0,0 +1,559 @@
...
+            "Could not extract temporal bounds from ERA5-WRF coverage metadata"
+        )
+
+
+def validate_era5_date_range(start_date, end_date, coverage_meta):
```
brucecrevensten, line 131: Remove

routes/era5wrf.py
```diff
@@ -0,0 +1,559 @@
...
+        for variable, values in data_dict.items():
+            if i < len(values) and values[i] is not None:
+                # round to 1 decimal good for current variable set, might need to change later
+                packaged_data[date_key][variable] = round(values[i], 1)
+
+    return packaged_data
+
+
+@routes.route("/era5wrf/")
+@routes.route("/era5wrf/abstract/")
```
brucecrevensten, line 219: We might want to make a ticket here to remove the `/abstract` endpoints throughout, that feels very early-days here I think?  (I know: I suggested it!  Not useful anymore?)

routes/era5wrf.py
```diff
@@ -0,0 +1,559 @@
...
+        dict: Variable names mapped to xarray datasets
+    """
+    time_slice = create_era5_time_slice(start_date, end_date)
+    bbox_bounds = polygon.total_bounds  # (xmin, ymin, xmax, ymax)
+
+    tasks = []
+    for var_name in variables:
+        cov_id = era5wrf_coverage_ids[var_name]
+
+        if time_slice:
```
brucecrevensten, line 277: Nuke

brucecrevensten (changes requested):

routes/era5wrf.py
charparr: ticketed in #618

BobTorgerson (commented):
All of the lat / lon and area summaries I tried came back with valid data and I tested the  ability to make CSVs with a variety of variables. This all looks great to me!

I added a few things that I think would be good to add. The only thing I'm not sure about is if we should hold off on merging this into main until we have a valid citation for the documentation rather than a placeholder text.

templates/documentation/era5wrf.html
```diff
@@ -0,0 +1,230 @@
...
+  for that date.
+</p>
+
+<h3>References and Data Sources</h3>
+
+<h3>Citing This Dataset</h3>
+
+<p>
+  When using this dataset, please cite the original ERA5 reanalysis and
+  acknowledge the WRF downscaling effort: [Citation details to be provided]
```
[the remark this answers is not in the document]
charparr: Good catch Bob. I'm going to add a citation for 
`
Waigl, C. F., Bieniek, P., Bhatt, U. S., Lader, R., & Walsh, J. (2024, January). A High-Resolution, Dynamically Downscaled ERA5 Reanalysis Dataset for Alaska. In 104th Annual AMS Meeting 2024 (Vol. 104, p. 439976).
`
This is the only citation - it is just for a conference proceeding, but that stuff is citable and there is an abstract / poster for people to look at.

I'm not sure if we want to wait until the journal article for this paper is out. As far as I know it is in prep, but not submitted, and so could be a year out. Hard to say. @brucecrevensten thoughts?

routes/era5wrf.py
```diff
@@ -0,0 +1,340 @@
...
+        reference_meta = era5wrf_meta[
+            "t2_mean"
+        ]  # any coverage metadata for time axis, they are all the same
+        packaged_data = package_era5wrf_point_data(all_data, reference_meta)
+        postprocessed = prune_nulls_with_max_intensity(
+            postprocess(packaged_data, "era5wrf_4km")
+        )
+
+        if request.args.get("format") == "csv":
+            return create_csv(postprocessed, "era5wrf_4km", lat=lat, lon=lon)
```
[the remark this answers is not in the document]
charparr: Thanks Bob! That pattern was not on my radar, I wouldn't have got there.

PR #623, opened 2025-08-20
Adjust last accessed to match NCR

### LINKED ISSUE #579: Update access dates in Demographics endpoint citations
It looks like the access dates of the following datasets got updated in NCR but not the Data API:

- 2020 US Census Demographics and Housing Characteristics File (DHC)
- 2023 CDC PLACES dataset

NCR says they were last accessed November, 2024. The Data API says they were accessed May, 2024. I'm sure they are the same version of the dataset either way, but since NCR uses the Data API, we may want to update the dates on the Data API side too.

brucecrevensten:
Closes #579 

Test by checking out the documentation endpoint for Demographics.  Ensure that the two datasets referenced in 579 say November and not May.


Joshdpaul (approved):
Looks good!

PR #636, opened 2025-09-23
CMIP6 Fire weather (CFFDRS) endpoint

Joshdpaul:
This PR adds an endpoint for the CMIP6 Fire Weather Indices dataset, aka "CFFDRS". (Read the EDA and learn more about the processing of that dataset here: https://github.com/ua-snap/ardac/tree/cmip6_fwi/curation/cmip6_fwi). The intention here is to prove that we can: 
- ingest daily fire weather data successfully
- query that data and use it to make data viz similar to the CATS data release
- provide a starting point for discussing possible "pots of gold" with fire people

### Rasdaman coverages with CF-compliant time dimension

Daily values for fire weather variables have been ingested into the following (more or less CF-compliant) Rasdaman coverages:

```
cmip6_bui
cmip6_dc
cmip6_dmc
cmip6_ffmc
cmip6_fwi
cmip6_isi
```

Only dates from April thru October are included in the coverages (just the fire season) and the dates are integer "days since 1980-04-01". This requires some additional functions to translate a standard request year (like 2020) to an integer value that matches the units of the coverage. For this, we use the coverage metadata to help parse requests and perform validation. When returns are encoded as netCDF, we don't need to do anything special - `xarray.open_dataset()` is able to read the metadata for the time dimension and automatically open the datasets with dates decoded into human-readable format.

Coverages were ingested using the following tiling recipe:

`"ALIGNED [0:3, 0:*, 0:1, 0:1] tile size 1643456"`

The idea here is to prioritize the time dimension, since we are grabbing large chunks of daily data over small areas. The tile size is based on the following math:

```
float32 = 4 bytes
models * timesteps * lon * lat = 4 x 25679 * 2 * 2  = 410864 values
410864 * 4 = 1643456 bytes = 1.64MB
```

### Custom aggregation: Date range and operations

One of the main benefits of ingesting daily data is the ability to perform custom aggregations of that data, based on user-provided parameters (otherwise, it would be easier to just pre-aggregate the data ourselves as we have done in the past). 

In this endpoint, I took a stab at this by allowing the user to input start and end years (i.e. define their own era buckets) and choose from a selection of aggregation operations to perform on the dataset. In the future, these options might change - it all depends on what users want to do with the data. Some of the operations could even be done via WCPS.

### Why these operations?

The choice of a 3-day rolling average + aggregation for each DOY across the time range was chosen because it was used to create this figure type in the [CATS data release](https://www.sciencebase.gov/catalog/item/686ec34bd4be020e5c0a6b74) - for example:
image

The choice of summer fire danger rating adjectives and the method of counting annual day classifications and averaging over the time range was chosen because it was used to create both figures and tables in the same CATS data release:
image

The actual adjective classes come from here:https://fire.ak.blm.gov/content/FuelFire/Alaska%20Fire%20Danger%20Operating%20Plan/Alaska_FDOP_FINAL_SIGNED.pdf:

image

## How to test:

You will need to rebuild your `pip` or `conda` virtual environment in order to add `cftime`.

Start the flask application as usual, and test some working endpoints:

- 3 day rolling average for select variables, select years (this is the default operation): http://localhost:5000/fire_weather/point/65.06/-146.16/2030/2050?vars=bui,fwi

- 5 day rolling average for select variables, all years: http://localhost:5000/fire_weather/point/65.06/-146.16?vars=bui,fwi&op=5_day_rolling_average

- average count of summer fire danger rating days for select variables, select years:
http://localhost:5000/fire_weather/point/65.06/-146.16/2030/2050?vars=bui,dmc,ffmc&op=summer_fire_danger_rating_days

- average count of summer fire danger rating days for select variables, all years:
http://localhost:5000/fire_weather/point/65.06/-146.16?vars=bui,dmc,ffmc&op=summer_fire_danger_rating_days

Replace any of the points above with a place ID to get zonal stats, for example:

- http://localhost:5000/fire_weather/area/1908030906

Use the `format=csv` parameters to get a CSV, for example:

- http://localhost:5000/fire_weather/area/1908030906/2040/2060?format=csv
- http://localhost:5000/fire_weather/area/1908030906/2040/2060?op=summer_fire_danger_rating_days&format=csv

Check out what happens to the `era5` model when we only query historical years (ie <=2021) versus when we query a time range that overlaps historical and projected. Any aggregation that overlaps between historical and projected can only use historical values for `era5`, so those value only cover a portion of the user-defined time range. In these cases, the `era5` values don't change as you move outside the historical era:
- 3 day rolling average for select variables, select years:
http://localhost:5000/fire_weather/point/65.06/-146.16/1990/2021?vars=fwi
http://localhost:5000/fire_weather/point/65.06/-146.16/1990/2030?vars=fwi

- average count of summer fire danger rating days for select variables, select years:
http://localhost:5000/fire_weather/point/65.06/-146.16/2000/2021?vars=bui&op=summer_fire_danger_rating_days
http://localhost:5000/fire_weather/point/65.06/-146.16/2000/2030?vars=bui&op=summer_fire_danger_rating_days

Note that the data domain is limited to the North American boreal zone (see here for a domain map: https://docs.google.com/document/d/16NsD-0LXwuXAS6jJT3NEx3Bv7CKeQB-QbGS_wAocMjs/edit?tab=t.0#heading=h.tmd0nil9wwib). When picking points / polygons for testing, anything outside of this domain should return 404. For example:

- http://localhost:5000/fire_weather/area/19020203 < HUC8 for Prince William Sound
- http://localhost:5000/fire_weather/point/70/-148 < point near Prudhoe Bay

### Questions to ponder / consider with collaborators:

- What other operations should we incorporate, if any? Do moving windows of 3, 5, and 7 days really matter? Should we consider a "raw" operation that just returns a big dump of all daily data? 
- Should user-defined parameters focus more on the time buckets (selecting year ranges and maybe specific months), or more on providing inputs to more complex functions (rolling aggregations over n-days, counting values above/below thresholds, conditional statements)?
- Do we really need all of the variables? It's unclear what fire folks actually use. Some variables can be combined to produce additional ratings: for instance, this GAR rating which is based off `isi` and `ffmc`:
image
the GAR table comes from here: https://fire.ak.blm.gov/content/FuelFire/Alaska%20Fire%20Danger%20Operating%20Plan/Alaska_FDOP_FINAL_SIGNED.pdf

### DIFF
--- a/application.py
+++ b/application.py
@@ -7,8 +7,8 @@ from config import SITE_OFFLINE, geojson_names
 from marshmallow import Schema, fields, validate, ValidationError
 import re
 
-from luts import fire_weather_ops
 from luts import (
+    fire_weather_ops,
     all_cmip6_downscaled_models,
     all_cmip6_downscaled_scenarios,
 )
@@ -114,9 +114,6 @@ def validate_get_params():
             required=False,
         )
 
-        # Make sure "op" parameter is one of the predefined fire weather operations
-        op = fields.Str(
-            validate=validate.OneOf(fire_weather_ops),
         # Make sure "vars" parameter is only letters and commas, and less than
         # or equal to 100 characters long.
         def validate_vars(value):
@@ -150,6 +147,12 @@ def validate_get_params():
 
         scenarios = fields.Str(
             validate=validate_scenarios,
+            required=False,
+        )
+
+        # Make sure "op" parameter is one of the predefined fire weather operations
+        op = fields.Str(
+            validate=validate.OneOf(fire_weather_ops),
             required=False,
         )
 


Joshdpaul:
Update: the `cmip6_isi` coverage in Rasdaman now works as it should, after re-ingesting with a different "description" variable attribute. Still no idea _why_ that was causing a problem, but its been fixed!

Pipfile
charparr: I think the in-flight PR #650 might end up stripping out all Pip-related stuff. I don't think you need to modify anything here, but we'll want to make sure we retain `cftime`!

routes/fire_weather.py
```diff
@@ -0,0 +1,753 @@
...
+        "end_date": cftime_value_to_ymd(
+            time_max, base_date
+        ),  # (year, month, day) tuple
+    }
+
+
+#### VALIDATION FUNCTIONS ####
+
+
+def validate_latlon(lat, lon):
```
charparr: I like this wrapper, but there is a function with this exact name in `validate_request.py` so this has some potential to be a trap. I'd suggest renaming this function and/or pushing this to the validate_request module where other endpoints could use this wrapper.

routes/fire_weather.py
```diff
@@ -0,0 +1,753 @@
...
+    if latlon_is_numeric_and_in_geodetic_range(lat, lon) == 400:
+        return render_template("400/bad_request.html"), 400
+    if check_geotiffs(float(lat), float(lon), coverages=[fire_weather_geotiff]) == 404:
+        print("no data at lat/lon")
```
charparr: Suggest logging this rather than printing

routes/fire_weather.py
```diff
@@ -0,0 +1,753 @@
...
+        url += f"&RANGESUBSET={var}"
+
+        tasks.append(fetch_bbox_netcdf([url]))
+
+    datasets = await asyncio.gather(*tasks)
+    datasets_dict = {var_name: ds for var_name, ds in zip(requested_vars, datasets)}
+    return datasets_dict
+
+
+def calculate_zonal_stats(polygon, datasets_dict, variables):
```
charparr: I suggest renaming this function as well to avoid collision with the function of the same name in `zonal_stats.py`

routes/fire_weather.py
```diff
@@ -0,0 +1,753 @@
...
+            return render_template("404/no_data.html"), 404
+        return render_template("500/server_error.html"), 500
+
+    n = None
+    if "3_day_rolling_average" in requested_ops:
+        n = 3
+    elif "5_day_rolling_average" in requested_ops:
+        n = 5
+    elif "7_day_rolling_average" in requested_ops:
+        n = 7
```
charparr: This gets repeated. I'd suggest making this a dict that is constant at the top of this file, or, dunking it in `luts.py`.

charparr (changes requested):
Yo this is an impressive endpoint Josh! There is a lot of work here, and I think this is going to be a data service that people will be excited about. I've left a few comments, most of them minor. Normally I wouldn't quibble about the naming of things, but in this instance I do think it is good hygiene to disambiguate further.

routes/fire_weather.py
brucecrevensten: Or nuke if this was just for dev reasons!

application.py
```diff
@@ -116,6 +115,14 @@ class QueryParamsSchema(Schema):
             required=False,
         )
 
+        # Make sure "op" parameter parameter is only letters, numbers, and commas, and less than
```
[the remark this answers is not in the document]
brucecrevensten: I agree, these are effectively enumerations / enums, treat them that way here and require they belong to an explicit list!

Pipfile
brucecrevensten: Yeah suggest rebase after 650 is merged.

brucecrevensten (changes requested):

luts.py
```diff
@@ -560,3 +563,49 @@
...
+        "Ext": (100, np.inf),
+    },
+    "ffmc": {
+        "Low": (0, 59),
+        "Mod": (59, 75),
+        "High": (75, 85),
+        "VHigh": (85, 90),
+        "Ext": (90, np.inf),
+    },
+    "fwi": {
```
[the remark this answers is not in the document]
Joshdpaul: Wow good catch! Eagle eyes 👁️ 👁️

routes/fire_weather.py
Joshdpaul: fixed viz [8ca7a6a](https://github.com/ua-snap/data-api/pull/636/commits/8ca7a6a2d09d6909c2853e73a637e84a2fcc7700)

luts.py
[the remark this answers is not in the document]
Joshdpaul: fixed via https://github.com/ua-snap/data-api/pull/636/commits/8458f3feb4078e7b0e77a0a77c0fbcc8f4382cb0

routes/fire_weather.py
Joshdpaul: fixed via [8458f3f](https://github.com/ua-snap/data-api/pull/636/commits/8458f3feb4078e7b0e77a0a77c0fbcc8f4382cb0)

PR #653, opened 2025-10-23
Add CMIP6 downscaled daily endpoint

cstephen:
This PR implements a new CMIP6 downscaled daily endpoint that currently serves the tasmax, tasmin, and pr variables.

Because daily data for just these 3 variables (+ historical & 4 future scenarios) across all models represents over 3tb of data, I experimented with lots of different ways to ingest and query data in the most performant way, and ultimately landed on the approach implemented in this PR. The approach is to ingest each model/scenario/variable combo as its own coverage (with `[0:*, 0:31, 0:31]` tiling), so that each coverage hits a 15gb - 26gb sweet spot, and make it so that the Data API requests only the bare minimum over WCS based on what GET parameters are specified through the endpoint URL.

This means that if a user requests just one model/scenario/variable combination, the Data API can return what was requested in about ~3 seconds. I rigged up the corresponding data x-ray items in ARDAC Explorer to request only the data needed to display the current chart, so that each charts takes about ~3 seconds to populate even without Varnish caching.

I think that's about all you need to know to dig into this PR. Make sure the documentation page makes sense and that the endpoint behaves as expected when you request one model/scenario/variable combo or multiple model/scenario/variable combos at the same time (the new `models` and `scenarios` GET parameters support comma-separated values to specify multiples).

### DIFF
--- a/templates/documentation/cmip6_downscaled.html
+++ b/templates/documentation/cmip6_downscaled.html
@@ -4,9 +4,10 @@
 <p>
   This endpoint provides access to daily CMIP6 data that have been bias-adjusted and downscaled to 4km using <a
     href="/era5wrf">dynamically downscaled ERA5</a>
-  reference data. Select models and scenarios are available. Read more about our CMIP6 subset
-  <a href="https://arcticdatascience.org/item/story-arctic-climate-data-node">here</a>. This dataset spans years
-  1965&ndash;2100 and its extent covers most of Alaska and parts of western Canada. A 6-model average is also available.
+  reference data. Twelve models are available with up to four future scenarios available per model. This is a subset of
+  the model and scenario combinations described <a
+    href="https://arcticdatascience.org/item/story-arctic-climate-data-node">here</a>. A 6-model average is also
+  available. This dataset spans years 1965&ndash;2100 and its extent covers most of Alaska and parts of western Canada.
 </p>
 
 <h3>Service endpoints</h3>
@@ -97,7 +98,7 @@
     <td>EC-Earth3-Veg</td>
     <td>historical, ssp126, ssp370, ssp585</td>
     <td>pr, tasmax, tasmin</td>
-    <td>ssp126 scenario only is only available for the pr variable.</td>
+    <td>ssp126 scenario is only available for the pr variable.</td>
   </tr>
   <tr>
     <td>HadGEM3-GC31-LL</td>


application.py
```diff
@@ -12,10 +12,8 @@
 # Configure logging to emit to stdout
 logging.basicConfig(
     level=logging.INFO,
-    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
-    handlers=[
-        logging.StreamHandler(sys.stdout)
-    ]
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+    handlers=[logging.StreamHandler(sys.stdout)],
```
charparr: This is properly (i.e. `black`) formatted. I've seen diffs like this in a few PRs now that are indicative of editor config bounce. Your version is correct, but I've created a ticket XREF #654 https://github.com/ua-snap/data-api/issues/654

application.py
```diff
@@ -33,6 +31,7 @@ def get_service_categories():
     """
     return [
         ("CMIP6", "/cmip6"),
+        ("CMIP6, Downscaled", "/cmip6_downscaled"),
```
charparr: I agree with this routing: fundamentally a different resource than the coarser cmip6

application.py
```diff
@@ -116,6 +115,16 @@ class QueryParamsSchema(Schema):
             required=False,
         )
 
+        models = fields.Str(
```
charparr: Also agree with these being query params and not being built into the route

templates/documentation/cmip6_downscaled.html
```diff
@@ -0,0 +1,149 @@
+{% extends 'base.html' %} {% block content %}
+<h2>CMIP6, Downscaled</h2>
+
+<p>
+  This endpoint provides access to daily CMIP6 data that have been bias-adjusted and downscaled to 4km using ERA5
+  reference data. Select models and scenarios are available. Read more about our CMIP6 subset
```
charparr: Slight preference for stating the number of models and scenarios here, along with a phrase that conveys we don't have the superset product of all combinations: availability varies by model, etc.

templates/documentation/cmip6_downscaled.html
```diff
@@ -0,0 +1,149 @@
...
+<p>
+  Below is the list of CMIP6 variables that are found in this endpoint. For
+  complete variable definitions, see the official CMIP6 documentation
+  <a href="https://wcrp-cmip.org/cmip-model-and-experiment-documentation/">here</a>.
+</p>
+
+<dl>
+  <dt>pr</dt>
+  <dd>
+    Precipitation, including both liquid and solid phases (kg m<sup>-2</sup> s<sup>-1</sup>)
```
charparr: Double check the units here vs. what is the CSV output, etc.

templates/documentation/cmip6_downscaled.html
```diff
@@ -0,0 +1,149 @@
...
+      <th>Metadata & source data access</th>
+      <th>Academic reference</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>
+        Publication in progress. For more information,
+        please
+        <a href="mailto:uaf-snap-data-tools@alaska.edu">contact us</a> directly.
```
charparr: I like this approach - gets the data out ahead of the publication, but if someone is hung up on that they are invited to send us an email 👍

templates/documentation/cmip6_downscaled.html
```diff
@@ -0,0 +1,149 @@
...
+  <tfoot>
+    <tr>
+      <td colspan="2">
+        Multiple models can be requested by appending a comma-separated string
+        like <code>?models=&lt;model_1&gt;,&lt;model_2&gt;</code>.<br />
+        Multiple scenarios can be requested by appending a comma-separated
+        string like <code>?scenarios=&lt;scenario_1&gt;,&lt;scenario_2&gt;</code>.<br />
+        Multiple variables can be requested by appending a comma-separated
+        string like <code>?vars=&lt;var_id_1&gt;,&lt;var_id_2&gt;</code>.<br />
+        CSV output is also available by appending <code>?format=csv</code> to
```
charparr: Add sentence(s) somewhere about how to get all vars and/or all models and/or all scenarios (i.e. seems like this happens by default when query params are omitted) which I think is a sane default: you don't know which model(s) you want? Then you get all of them.

routes/cmip6_downscaled.py
```diff
@@ -0,0 +1,262 @@
...
+all_possible_vars = list(cmip6_downscaled_options.keys())
+all_possible_models = set(
+    model for var in cmip6_downscaled_options.values() for model in var.keys()
+)
+all_possible_scenarios = set(
+    scenario
+    for var in cmip6_downscaled_options.values()
+    for model in var.values()
+    for scenario in model
+)
```
charparr: This could get moved to `luts.py` if it appeals to your taste

routes/cmip6_downscaled.py
```diff
@@ -0,0 +1,262 @@
...
+    Args:
+        metadata (dict): coverage metadata
+        point_data_list (list): list of data from Rasdaman WCPS query
+
+    Returns:
+        di (dict): time series dictionary of date/value pairs
+    """
+    di = dict()
+
+    time_series = generate_time_index_from_coverage_metadata(metadata)
```
charparr: This feels very pleasant - tidy time-stamping of the package.

routes/cmip6_downscaled.py
```diff
@@ -0,0 +1,262 @@
...
+    Args:
+        cov_id (str): coverage ID
+        x (float): x-coordinate
+        y (float): y-coordinate
+
+    Returns:
+        list of data results at a specified point
+    """
+
+    # We must use EPSG:4326 for the CMIP6 downscaled coverage to match the coverage projection
```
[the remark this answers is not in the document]
cstephen: Whoops, this comment was copied from the non-downscaled CMIP6 endpoint and never got updated. The line of code below it is simple and self-explanatory IMO, so I'm just going to remove this comment.

routes/cmip6_downscaled.py
cstephen: Fixed in fdfb744.

routes/cmip6_downscaled.py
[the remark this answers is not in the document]
cstephen: Fixed in fdfb744.

templates/documentation/cmip6_downscaled.html
```diff
@@ -0,0 +1,149 @@
...
+        <a href="/cmip6_downscaled/point/61.5/-147?models=6ModelAvg&scenarios=ssp585&vars=tasmax">
+          /cmip6_downscaled/point/61.5/-147?models=6ModelAvg&scenarios=ssp585&vars=tasmax</a>
+      </td>
+    </tr>
+  </tbody>
+  <tfoot>
+    <tr>
+      <td colspan="2">
+        Multiple models can be requested by appending a comma-separated string
+        like <code>?models=&lt;model_1&gt;,&lt;model_2&gt;</code>.<br />
```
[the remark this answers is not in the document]
Joshdpaul: Agreed! Consider linking to here: http://127.0.0.1:5000/cmip6/references 
Or even creating a small table like so:

| Model | Scenarios | Ensemble Member |
| --- | --- | --- | 
| CESM2 | historical, ssp126, ssp245, ssp370, ssp585 | ❌ |
| ... | ... | ... |
| TaiESM1 | historical, ssp126, ssp245, ssp370, ssp585 | ✅ |

Joshdpaul (commented):
Great job with this Craig! It's a tidy endpoint, and performs well given the large number of coverages and the large size of the dataset. Looks like it will be easily expandable to more variables if/when we downscale more data. 

Speaking of that, I don't know if its appropriate for this branch now or would be better considered in another branch, but I am wondering if we should include a `tas` variable here? I realize that the coverages are just `tasmin` and `tasmax`, but it would not be a huge lift to average these together in the API and provide a daily average temperature variable. This could be accomplished by querying both `tasmin` and `tasmax` using the existing code here, and then adding a post-processing function to get the mean for each date and return a new dictionary. Alternatively, we could write a new WCPS query to do this for us. The first option is probably the easiest thing to implement.

Otherwise, beyond the slight improvements to the documentation that Charlie suggested, I think this is looking fantastic!

templates/documentation/cmip6_downscaled.html
cstephen: Good call here, and fixed in a4f4ad5 to be "mm/day" (to be precise). The CSV metadata line says:

> pr is the daily precipitation in mm.

Which is a more verbose way to say the same thing, but they are both in agreement now.

templates/documentation/cmip6_downscaled.html
cstephen: Fixed in 9de32ba, I think!

charparr (approved):
Craig this is looking great! I think this endpoint is strong, and the doc tweaks nicely capture some of the idiosyncrasies in terms of the data we have available, and should be a handy reference, both internal and external, for "what do we offer for cmip6 downscaled?"

Joshdpaul (approved):

PR #667, opened 2025-12-01
Adds an API endpoint to retrieve the landslide risk data for the ḵutí project

BobTorgerson:
This PR adds a new API endpoint to retrieve the landslide risk data for the ḵutí project website. To use this branch, we need to provide the read-only credentials of the web user in the Keeper item: Landslide Database Read-Only User

The following environment variables need to be populated for this to run:
**DB_HOST, DB_NAME, DB_USER, and DB_PASSWORD**

The DB_NAME is landslide_risk and the DB_USER is web. The rest can be obtained from the Keeper item.

This will add the ability to access API output for the URL: http://localhost:5000/landslide/Craig or http://localhost:5000/landslide/Kasaan

### DIFF
--- a/routes/landslide.py
+++ b/routes/landslide.py
@@ -115,6 +115,4 @@ def run_fetch_landslide_data(community_id):
 
     except Exception as exc:
         logger.error(f"Error in landslide endpoint for {community_id}: {exc}")
-        if hasattr(exc, "status") and exc.status == 404:
-            return render_template("404/no_data.html"), 404
         return render_template("500/server_error.html"), 500


routes/landslide.py
```diff
@@ -0,0 +1,77 @@
...
+
+    try:
+        results = get_landslide_db_row(place_name)
+        landslide_data = package_landslide_data(results)
+        return landslide_data
+
+    except Exception as exc:
+        if hasattr(exc, "status") and exc.status == 404:
+            return render_template("404/no_data.html"), 404
+        return render_template("500/server_error.html"), 500
```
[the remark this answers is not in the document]
brucecrevensten: Add a single basic e2e integration test for this?  Why not.

brucecrevensten (changes requested):
Suggest,
 - Add a basic docs page that links out to the repo that defines the data structure (no need to get fancy, but could show a data return example) and the upstream data sources.
 - Add to catalog page.
 - Use the community IDs we normally use and bundle the place names in the JSON response + lat/lng + all the info we have for them.  This is more consistent, even if we don't expand beyond these few communities.
 - (Consider) add link to Readme that also x-ref's the repo defining the data structure
 - Add a basic e2e test that does a fetch + tests for existence

fetch_data.py
```diff
@@ -501,3 +557,49 @@ def cftime_value_to_ymd(time_value, base_date):
     """Convert a time value in days since the base date to a year, month, day tuple."""
     date = base_date + datetime.timedelta(days=time_value)
     return date.year, date.month, date.day
+
+
+def get_place_data(place_id):
```
brucecrevensten: I think here we can lean on the `vectordata.py` route which looks like it already has this functionality baked in.  Could a simple function be added there, "get_community_by_id" that returns whatever is present in the `all_communities_full` variable that looks like it's populated at startup?

routes/landslide.py
```diff
@@ -0,0 +1,102 @@
...
+import logging
+
+from . import routes
+from fetch_data import get_landslide_db_row, get_place_data
+from validate_data import place_name_and_type
+
+logger = logging.getLogger(__name__)
+
+
+def validate_community_id(community_id):
```
brucecrevensten: I think to keep this really simple you can just test the supplied strings.  We shouldn't have translation from upper/lowercase unless things are manually entered, which is sort of not a supported use case right?  And it'd be great to untangle this thread of multiple code blocks that do the same WFS request.

routes/landslide.py
```diff
@@ -0,0 +1,127 @@
...
+        "precipitation_inches": data.get("precip_inches"),
+        "precipitation_24hr": data.get("precip24hr"),
+        "precipitation_2days": data.get("precip2days"),
+        "precipitation_3days": data.get("precip3days"),
+        "risk_level": data.get("risk_level"),
+        "risk_probability": data.get("risk_prob"),
+        "risk_24hr": data.get("risk24hr"),
+        "risk_2days": data.get("risk2days"),
+        "risk_3days": data.get("risk3days"),
+        "risk_is_elevated_from_previous": data.get("risk_is_elevated_from_previous"),
```
brucecrevensten: Remove

validate_request.py
```diff
@@ -254,17 +253,11 @@ def validate_var_id(var_id):
...
 
-    var_id_check = asyncio.run(
-        fetch_data(
-            [generate_wfs_places_url("all_boundaries:all_areas", "type", var_id, "id")]
-        )
-    )
-
-    if var_id_check["numberMatched"] > 0:
-        return var_id_check["features"][0]["properties"]["type"]
+    for area in all_areas_full:
```
brucecrevensten: Use a direct dict lookup after modifying the data structure.

fetch_data.py
```diff
@@ -501,3 +564,50 @@ def cftime_value_to_ymd(time_value, base_date):
...
+    fetch_data(
+        [
+            generate_wfs_places_url(
+                "all_boundaries:all_communities",
+                "name,alt_name,id,region,country,type,latitude,longitude,tags,is_coastal,ocean_lat1,ocean_lon1",
+            )
+        ]
+    )
+)["features"]
+
```
brucecrevensten: Now reroll them so it's a dict with key-pairs of IDs for fast lookup.

tests/test_landslide.py
```diff
@@ -0,0 +1,51 @@
+import json
+
+
+def test_landslide_ak91(client):
```
brucecrevensten: Adjust per conversations

tests/test_landslide.py
```diff
@@ -0,0 +1,51 @@
...
+    assert response.status_code == 200
+    actual_data = response.get_json()
+
+    with open("tests/landslide_risk_AK182.json") as f:
+        expected_data = json.load(f)
+
+    assert actual_data == expected_data
+
+
+def test_landslide_invalid_community(client):
```
brucecrevensten: Good

tests/test_landslide.py
```diff
@@ -0,0 +1,51 @@
...
+    assert response.status_code == 200
+    actual_data = response.get_json()
+
+    with open("tests/landslide_risk_AK91.json") as f:
+        expected_data = json.load(f)
+
+    assert actual_data == expected_data
+
+
+def test_landslide_ak182(client):
```
brucecrevensten: Remove, just replace this one with the monkeypatched stuff I think

tests/landslide_risk_AK182.json
```diff
@@ -0,0 +1,30 @@
+{
```
brucecrevensten: Remove per comment below

templates/documentation/landslide.html
```diff
@@ -0,0 +1,145 @@
...
+  "precipitation_inches": 0.49,
+  "precipitation_24hr": 45.2,
+  "precipitation_2days": 78.1,
+  "precipitation_3days": 95.6,
+  "risk_level": 2,
+  "risk_probability": 0.25,
+  "risk_24hr": 3,
+  "risk_2days": 4,
+  "risk_3days": 4,
+  "risk_is_elevated_from_previous": true,
```
brucecrevensten: Remove

brucecrevensten (changes requested):
There may be some dupe comments.

routes/landslide.py
```diff
@@ -0,0 +1,127 @@
...
+    Validate that the community ID is both a valid place ID and one of the supported landslide locations.
+    Uses the existing place validation system, then restricts to AK182 and AK91.
+
+    Args:
+        community_id (str): The community ID to validate (AK182 for Kasaan, AK91 for Craig)
+
+    Returns:
+        str or None: Place name if valid community ID, None if invalid
+    """
+    place_name, place_type = place_name_and_type(community_id)
```
brucecrevensten: Can you change this JUST to be a string comparison?

Issue #682, opened 2026-01-13
Add water year index to `conus_hydrology/modeled_climatology`


Joshdpaul:
Hydrographs are commonly displayed using the water year (Oct 1 - Sept 30). 

Our current `conus_hydrology/modeled_climatology` API return gives the daily climatology min/mean/max for each day of year, but this creates problems for indexing the values on the hydrograph x-axis and for applying smoothing to the time series. 

For our dataset's 366-day calendar, the axis values for the water year would run from DOY 275-366, then from 1-274. Our return should also provide a "water year index" value from 1-366, to make hydrograph visualizations easier to code on the front end.

PR #684, opened 2026-01-15
Add water year index

Joshdpaul:
This is a minor feature addition, that calculates the water year index (Oct 1 = 1, Sep 30 = 366) for each day of year and includes it as part of the return for `conus_hydrology/modeled_climatology` and `conus_hydrology/observed_climatology`. Closes https://github.com/ua-snap/data-api/issues/682

The idea here is to remove code the front end for plotting hydrographs, and/or make it easier for API users to plot their own hydrographs.

To test:

Check out returns for `conus_hydrology/modeled_climatology` and `conus_hydrology/observed_climatology` endpoints, and make sure the `water_year_index` values make sense. Also request some CSVs, and make sure the metadata and `water_year_index` values appear as they should.

### DIFF
--- a/routes/conus_hydrology.py
+++ b/routes/conus_hydrology.py
@@ -160,7 +160,6 @@ async def get_usgs_gauge_data(gauge_id):
         df["discharge_cfs"] = values_df["discharge_cfs"]
 
     df["DOY"] = df.index.dayofyear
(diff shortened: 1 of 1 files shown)


brucecrevensten:
Looks good!

PR #688, opened 2026-01-21
Updates TAS 2km endpoint /tas2km/point/ to use the new WCS-optimized Rasdaman coverages.

BobTorgerson:
This PR changes the /tas2km/point/ endpoint to use the new coverages that have been uploaded to Zeus called tas_2km_historical_wcs and tas_2km_projected_wcs. These two coverages have been modified to use improved tiling to allow for much faster response times for the endpoint. From my initial testing, the new coverages respond in 0.5% of the original time going from a 3.9 minute wait to a 1.4 second response. 

For testing, you can swap the coverage names in the code below back to the original coverages and back to the new coverages to get the timing for both in the /tas2km/point/ endpoint.

Example URL: http://localhost:5000/tas2km/point/65.0628/-147.1627


cstephen (changes requested):
I compared the new coverages vs. the old (both running locally, by changing the coverage names in the API code). The performance difference is unreal!

I chose five lat/lon points in and around Alaska and Canada, downloaded the JSON for each on new vs. old coverages, and did a `diff` to confirm that the output is exactly the same. Four out of the five lat/lon points had identical outputs vs. the old API coverages. The one point that's different is this:

http://localhost:5000/tas2km/point/57.599933/-160.050385

Running the API using the old coverages returns a 404 error page template. The new coverages return a JSON object full of nothing but `null` values, which has downstream effects on applications like NCR. Seems like the coverages were ingested with a different nodata value than before or something. Might be fixable in the `nodata_values` / `nodata_mappings` dictionaries in the `postprocessing.py` file, though.

Overall, though, this is an incredible improvement for this endpoint!!

cstephen (approved):
I've confirmed that commit 067844e fixed http://localhost:5000/tas2km/point/57.599933/-160.050385 on my end, too. Thanks for the quick fix, @BobTorgerson!

PR #698, opened 2026-02-14
Improves zonal statistics for faster returns for area summaries

BobTorgerson:
This PR modifies the functions used to generate the zonal statistics, and the endpoints that call those functions. 

**Summary**
The reason for this change is that in my exploration into caching the area summaries, I noticed that while the Rasdaman queries were taking an average of 10 seconds to return, the zonal statistics were taking 30+ seconds to complete in the Python code after receiving the data from Rasdaman. As such, I decided to explore whether the zonal statistics computation could be improved to reduce that wait time for large polygons. The great news is that I found a simple thing that could be done to improve performance and we are seeing 2-4x speed improvements over previous endpoint calls with this new PR.

**What was done?**
I discovered that the function that was taking the longest to complete per iteration was interpolate(). Previously, we would iterate through all dimension combinations sequentially, sending each to the interpolate_and_compute_zonal_stats() function one by one. This would mean that we had to interpolate the data for those singular dimensions and rasterize the polygon for each iteration of dimensions, leading to work being done hundreds of times that could instead be done a single time. 

As such, the code now sends the entirety of the dataset returned from Rasdaman to the function and interpolates and rasterizes the polygon a single time. 

**Improvements**
With this change, I'm seeing approximately 2x speed up of the /taspr/ and /indicators/cmip5/ endpoint area summaries and a 4x speed up of the /alfresco/ endpoint area summaries.

**Testing**
To test these changes, you can time the runs in the Network tab of your browser for your local API on this branch vs the API running without a cache at: http://data-api-prod-2026.us-west-2.elasticbeanstalk.com

Examples to try:

ALFRESCO Flammability
http://localhost:5000/alfresco/flammability/area/19080309  **7.57 seconds**
http://data-api-prod-2026.us-west-2.elasticbeanstalk.com/alfresco/flammability/area/19080309 **24.37 seconds**

ALFRESCO Vegetation Type
http://localhost:5000/alfresco/veg_type/area/19080309  **11.66 seconds**
http://data-api-prod-2026.us-west-2.elasticbeanstalk.com/alfresco/veg_type/area/19080309 **45.69 seconds**

CMIP5 Indicators
http://localhost:5000/indicators/cmip5/area/1903040601/ **4.81 seconds**
http://data-api-prod-2026.us-west-2.elasticbeanstalk.com/indicators/cmip5/area/1903040601/ **9.24 seconds**

Temperature and Precipitation
http://localhost:5000/taspr/area/19010208 **9.62 seconds**
http://data-api-prod-2026.us-west-2.elasticbeanstalk.com/taspr/area/19010208 **26.64 seconds**

Please do try a variety of different area types and confirm that the data between the old and new version of the API match in output. Everything should be the same since none of the processing code was affected by this change as this was largely an efficiency change.

### DIFF
--- a/tests/test_indicators.py
+++ b/tests/test_indicators.py
@@ -3,10 +3,10 @@ import json
 
 def test_indicators_cmip5_area(client):
     """
-    Tests the /indicators/cmip5/area/<id> endpoint to ensure the output
+    Tests the /indicators/cmip5/area/<id>/ endpoint to ensure the output
     remains consistent with production for the given area ID.
     """
-    response = client.get("/indicators/cmip5/area/1903040601")
+    response = client.get("/indicators/cmip5/area/1903040601/")
     assert response.status_code == 200
     actual_data = response.get_json()
 


charparr (approved):
Good boost here! The size-of-diff to performance boost ratio here is probably hard to beat! I've added tests. And these tests now run faster too 😎

PR #709, opened 2026-03-19
Add all vs. mid (rcp60) outputs for hydrograph and monthly flow charts, and min/mid/max for table stats for hydroviz webapp

cstephen:
This is a tentative implementation of separate min/max and middle-of-the-road (RCP 6.0) outputs for hydroviz hydrographs, mean monthly flow charts, and table stats. I'm going to put this in draft mode until we verify that this has everything we need on the webapp side.


cstephen:
This PR is stale and has been replaced by similar work in #715.

PR #714, opened 2026-04-09
Removes GeoTIFF mask for HSIA production coverage

BobTorgerson:
This PR removes the GeoTIFF mask for the hsia_arctic_production Rasdaman coverage as it was masking out valid pixels of communities on the shoreline in our https://seaiceatlas.org website. 
Screenshot 2026-04-09 at 10 17 18 AM

You can see in the picture above that anywhere that color is being shown through the mask is a valid pixel that was incorrectly being marked as having no data. The code that was being used for this is in the check_geotiff function found here: https://github.com/ua-snap/data-api/blob/3b919c03b1e1de21a0a815c143e6c7cfb62289ae/validate_request.py#L73-L121

**NOTE:** When this goes into production, we are going to want to clear the Earth Maps cache of the /seaice endpoint so we can get all valid data back again.


charparr (approved):
Image

Easy to test with a side-by-side requests for a location where we'd expect data, but we don't get it in prod: https://seaiceatlas.org/report/58.760/-69.500#report

PR #737, opened 2026-07-31
Fix arctic_hydrology climatology Blaskey-adjusted JSON vs. original GCM CSV  discrepancies

cstephen:
This PR fixes the discrepancy in values between outputs of these two URLs (for example):

http://localhost:5000/arctic_hydrology/modeled_climatology/81009008
http://localhost:5000/arctic_hydrology/modeled_climatology/81009008?format=csv

The problem was that the `gcm_diff_applied_to_blaskey` adjustment was being done after the CSV export code, not before, for both the `modeled_climatology` and `wt_modeled_climatology` endpoints.

I've also added a couple chunks of code to move the "historical" model above the projected models in the CSVs to make them more intuitive and consistent with our other CSVs.

To test, I pointed the hydroviz webapp at this `json_csv_discrepancy` Data API branch and disabled LOWESS smoothing + enabled hovermode to cross reference values from the hydrographs & water temperature hydrographs with their corresponding CSVs to make sure that the same values appear on the same dates. I also cross-referenced a handful of flow & temperature stat table values against their corresponding CSVs. Everything is as expected!

### DIFF
--- a/routes/arctic_hydrology.py
+++ b/routes/arctic_hydrology.py
@@ -823,6 +823,14 @@ def run_get_arctic_hydrology_modeled_climatology(stream_id):
         data_dict = populate_feature_attributes(data_dict, gdf)
         data_dict = prune_nulls_with_max_intensity(data_dict)
 
+        # apply GCM-projected changes to Blaskey climatology if source is "gcm_diff_applied_to_blaskey"
+        # PGW models without a 1990-2021 historical era are silently absent from this source;
+        # they are only available via source=original_gcm.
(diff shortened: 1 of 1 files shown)


charparr (approved):
I've tested the JSON and CSV routes for a few different IDs, both CONUS and AK, and can verify they are aligned in terms or structure, numerical content, and associated metadata.

charparr:
Note the test failures here are just referencing missing env vars for kuti, so can disregard.

PR #739, opened 2026-07-31
Stabilize modeled-climatology change factors and fix CONUS CSV/JSON divergence

brucecrevensten:
## Background

Investigating `arctic_hydrology/modeled_climatology/81009008` revealed that the JSON output reported a `doy_min` of **602,159.949 cfs** for C2LE2 at doy 119 (2034–2065) — ~70× above its own `doy_max` — while the CSV export of the same endpoint showed entirely different numbers. Three distinct defects were isolated, all verified numerically against production:

1. **CSV/JSON divergence** — the CSV branch returns *before* the GCM-diff adjustment, so CSV serves raw original-GCM values under a header claiming they're Blaskey/Maurer-adjusted. @cstephen's `json_csv_discrepancy` branch fixes this for the two Arctic routes; **this PR intentionally leaves those hunks to that branch** and fixes only the CONUS route (no conflicts — different files/regions).
2. **Ratio blow-up near zero** — the change factor `gcm_projected / gcm_historical` explodes when the historical minimum is near zero (0.001 cfs winter flows on frozen streams). The one-sided `0 → 0.0001` floor makes this worse, not better. The 602,159.949 value is exactly `1.223 × (492.363 / 0.001)`.
3. **Internal inconsistency** — each statistic is scaled by its own independent ratio, so adjusted triplets can violate `min ≤ mean ≤ max` even at ordinary magnitudes. For CONUS stream 50563, 409 of 115,656 adjusted entries were inconsistent (raw data: 0), e.g. doy 214 CCSM4/rcp26: adjusted min 353.0 > adjusted max 343.7.

## Changes (one commit per fix)

**Commit 1 — CONUS: apply Maurer adjustment before CSV export.** Mirrors the Arctic fix on `json_csv_discrepancy`; both formats now serve the same adjusted data, matching the CSV header's description.

**Commit 2 — Stabilized change factors** (shared helper in `postprocessing.py`, used by both Blaskey and Maurer functions):

```
ratio = (projected + ε) / (historical + ε), clamped to [0.1, 10]
ε = 1% of the model's historical mean flow (floored at 0.0001)
```

- The symmetric ε makes the factor approach 1 ("no change") as both values approach zero, instead of diverging — and it's scale-aware, so it behaves the same for headwater creeks and large rivers.
- The `[0.1, 10]` cap is a backstop against noisy order statistics; ordinary ratios (e.g. mean flows on healthy baselines) are essentially unchanged (< 2% shift at 81009008).
- Effect at 81009008 doy 119: C2LE2 `doy_min` goes from 602,159.949 → ≈ 12.2 cfs.

**Commit 3 — Enforce `min ≤ mean ≤ max`** after adjustment: clamp `doy_min` down to `doy_mean` and `doy_max` up to `doy_mean`.

## Testing

- `tests/test_hydrology_adjustments.py` (14 tests): reproduces the production doy-119 blowup and the doy-214 ordering violation with real production values; covers epsilon scaling, clamping, PGW-model skipping, and structure preservation for both adjustment functions.
- Full suite: 33 passed; the 2 `test_landslide` failures pre-exist on `main` (require a live database).

## Notes for review

- ε fraction (1%) and cap (10×) are judgment calls — happy to tune. Rationale: the cap bounds fabricated values; ε controls how quickly small-flow ratios damp toward 1.
- Follow-up candidates (not in this PR): the CONUS JSON serves each model's `historical` scenario as an exact copy of Maurer post-adjustment (redundant payload); the water-temperature additive adjustment can also produce min/mean/max crossings; a methodologically stronger long-term fix is computing change factors on 31-day windows (or applying deltas to the daily series upstream in arctic_rivers) — see the investigation writeup for details.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/postprocessing.py
+++ b/postprocessing.py
@@ -267,3 +267,48 @@ def merge_dicts(dict1, dict2):
         else:
             merged[key] = value
     return merged
+
+
+# Stabilization constants for GCM change factors ("deltas") that are applied to
+# historical baselines as ratios. The epsilon fraction sets the offset relative
+# to the stream's historical mean flow; the floor keeps the offset nonzero for
+# all-zero streams; the cap bounds how much any single statistic can be scaled.
+RATIO_EPSILON_FRACTION = 0.01
+RATIO_EPSILON_FLOOR = 0.0001
+RATIO_CAP = 10.0
+
+
+def scale_aware_epsilon(values):
+    """
+    Compute a stabilizing offset for ratio-based change factors, sized to the
+    magnitude of the historical data so that the same code works for large
+    rivers and small creeks alike.
+    Args:
+        values (list of float): Historical baseline values (e.g. all doy_mean
+            values for one model), used to establish the stream's flow scale
+    Returns:
+        float: RATIO_EPSILON_FRACTION times the mean of values, but never less
+        than RATIO_EPSILON_FLOOR
+    """
+    if not values:
+        return RATIO_EPSILON_FLOOR
+    epsilon = RATIO_EPSILON_FRACTION * (sum(values) / len(values))
+    return max(epsilon, RATIO_EPSILON_FLOOR)
+
+
+def stabilized_ratio(projected, historical, epsilon):
+    """
+    Ratio of projected to historical, stabilized for near-zero baselines.
+    Adding the same epsilon to numerator and denominator makes the ratio
+    approach 1 (no change) as both values approach zero, instead of exploding
+    when only the denominator is small. The result is clamped to
+    [1/RATIO_CAP, RATIO_CAP] as a backstop against noisy statistics.
+    Args:
+        projected (float): Future (projected) statistic value
+        historical (float): Historical statistic value for the same model
+        epsilon (float): Positive stabilizing offset from scale_aware_epsilon()
+    Returns:
+        float: Clamped change factor suitable for scaling a historical baseline
+    """
+    ratio = (projected + epsilon) / (historical + epsilon)
+    return min(max(ratio, 1.0 / RATIO_CAP), RATIO_CAP)
--- a/routes/arctic_hydrology.py
+++ b/routes/arctic_hydrology.py
@@ -23,7 +23,11 @@ from generate_urls import (
 )
 from fetch_data import fetch_data, fetch_layer_data, describe_via_wcps
 from validate_request import get_axis_encodings
-from postprocessing import prune_nulls_with_max_intensity
+from postprocessing import (
+    prune_nulls_with_max_intensity,
+    scale_aware_epsilon,
+    stabilized_ratio,
+)
 from csv_functions import create_csv
 from config import RAS_BASE_URL
 from . import routes
@@ -490,6 +494,11 @@ def calculate_and_apply_gcm_diffs_to_blaskey_climatology(data_dict):
     """
     Function to calculate GCM-projected changes in streamflow and apply them to the historical Blaskey climatology.
     Models without a '1990-2021' era (e.g. PGW models with no historical baseline) are silently skipped.
+
+    Change factors are stabilized ratios (see postprocessing.stabilized_ratio):
+    a symmetric offset scaled to the model's historical mean flow keeps the
+    factor near 1 when both values are near zero (e.g. frozen winter minimums),
+    and factors are clamped to [1/RATIO_CAP, RATIO_CAP].
     Args:
         data_dict (dict): Climatology data dict keyed by model then era
     Returns:
@@ -502,6 +511,13 @@ def calculate_and_apply_gcm_diffs_to_blaskey_climatology(data_dict):
             continue
         if "1990-2021" not in data_dict[model]:
             continue
+        epsilon = scale_aware_epsilon(
+            [
+                row["doy_mean"]
+                for row in data_dict[model]["1990-2021"]
+                if "doy_mean" in row
+            ]
+        )
         adjusted_data_dict[model] = {}
         for era in data_dict[model].keys():
             if era == "1990-2021":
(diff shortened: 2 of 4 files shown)


cstephen:
I checked a bunch of arbitrary streams (for both CONUS and Alaska) from both futurehydrology.org and my local webapp, pointed at the `stabilize-climatology-ratios` Data API branch, to confirm that the new math improves the hydrographs. Indeed it does!

Here are some side-by-side URLs with old math vs. new math, and in every case it's an improvement:

https://futurehydrology.org/alaska/stream/81024981
http://localhost:3000/alaska/stream/81024981

https://futurehydrology.org/alaska/stream/81010040
http://localhost:3000/alaska/stream/81010040

https://futurehydrology.org/alaska/stream/81007312
http://localhost:3000/alaska/stream/81007312

https://futurehydrology.org/alaska/stream/81010509
http://localhost:3000/alaska/stream/81010509

https://futurehydrology.org/alaska/stream/81000040
http://localhost:3000/alaska/stream/81000040

https://futurehydrology.org/alaska/stream/81019526
http://localhost:3000/alaska/stream/81019526

https://futurehydrology.org/alaska/stream/81007411
http://localhost:3000/alaska/stream/81007411

https://futurehydrology.org/conus/stream/39046
http://localhost:3000/conus/stream/39046

https://futurehydrology.org/conus/stream/7532
http://localhost:3000/conus/stream/7532

https://futurehydrology.org/conus/stream/33308
http://localhost:3000/conus/stream/33308

https://futurehydrology.org/conus/stream/30118
http://localhost:3000/conus/stream/30118

https://futurehydrology.org/conus/stream/56453
http://localhost:3000/conus/stream/56453

I found one stream whose hydrographs got worse with the change, but after closer inspection, this turned out to be a problem with the webapp's client-side LOWESS smoothing, not the Data API data, and can be fixed by changing the LOWESS smoothing span (`f`) on the webapp side (with more testing on the webapp side of course):

https://futurehydrology.org/conus/stream/53799
http://localhost:3000/conus/stream/53799

postprocessing.py
```diff
@@ -267,3 +267,48 @@ def merge_dicts(dict1, dict2):
...
     return merged
+
+
+# Stabilization constants for GCM change factors ("deltas") that are applied to
+# historical baselines as ratios. The epsilon fraction sets the offset relative
+# to the stream's historical mean flow; the floor keeps the offset nonzero for
+# all-zero streams; the cap bounds how much any single statistic can be scaled.
+RATIO_EPSILON_FRACTION = 0.01
+RATIO_EPSILON_FLOOR = 0.0001
+RATIO_CAP = 10.0
```
charparr, line 278: One immediate question I have is how frequently are the floor / cap values triggered? If these values are returned frequently then they might need adjustment. It seems like they are intended to catch outliers which I think is the right approach but it could be worth an empirical test before all is said and done. I do see that Claude called these out already as "judgement calls" that Claude is "happy to tune".

postprocessing.py
```diff
@@ -267,3 +267,48 @@ def merge_dicts(dict1, dict2):
...
+    Args:
+        values (list of float): Historical baseline values (e.g. all doy_mean
+            values for one model), used to establish the stream's flow scale
+    Returns:
+        float: RATIO_EPSILON_FRACTION times the mean of values, but never less
+        than RATIO_EPSILON_FLOOR
+    """
+    if not values:
+        return RATIO_EPSILON_FLOOR
+    epsilon = RATIO_EPSILON_FRACTION * (sum(values) / len(values))
```
charparr, line 295: One choice I see embedded here is that the arithmetic mean is the value used to scale the epsilon. This makes sense to me because it is easy to explain and to conceptualize: big historical flows get bigger epsilons, and smaller historical flows get smaller epsilons.

But, are there any streamflow regimes for which this might have unanticipated consequences? Imagine a stream where the flow is: `[small, small, small, BIG BIG GUSH O WATER, small, small, ...]` : the mean may not scale the epsilon value in a desirable way because the big flow might get only get scaled a tiny bit.

Down the road it may be worthwhile exploring the use of the median the scale factor, or some other statistic. 

And/or just adding some more text along the lines of 
"Hey for a stream that is highly variable or seasonal, a spike in flow is going to raise the mean, which will raise the epsilon for the entire year's worth of flows.

routes/arctic_hydrology.py
```diff
@@ -519,12 +535,20 @@ def calculate_and_apply_gcm_diffs_to_blaskey_climatology(data_dict):
...
-                        denominator = 0.0001
-                    projected_quotient = gcm_projected / denominator
+                    projected_quotient = stabilized_ratio(
+                        gcm_projected, gcm_historical, epsilon
+                    )
                     blaskey_adjusted = round(blaskey_historical * projected_quotient, 3)
                     doy_stats[stat] = blaskey_adjusted
+                # each stat is scaled by its own ratio, so the adjusted values
+                # can cross; clamp to preserve min <= mean <= max
+                if all(k in doy_stats for k in ("doy_min", "doy_mean", "doy_max")):
```
charparr, line 545: Good comment on the "why" -- feels like we'll come back to this

routes/conus_hydrology.py
```diff
@@ -819,14 +835,24 @@ def calculate_and_apply_gcm_diffs_to_maurer_climatology(data_dict):
...
+                    # values can cross; clamp to preserve min <= mean <= max
+                    if all(
+                        k in doy_stats for k in ("doy_min", "doy_mean", "doy_max")
+                    ):
+                        doy_stats["doy_min"] = min(
+                            doy_stats["doy_min"], doy_stats["doy_mean"]
+                        )
+                        doy_stats["doy_max"] = max(
+                            doy_stats["doy_max"], doy_stats["doy_mean"]
+                        )
```
charparr, line 855: I think this is identical to R543 to R551 in routes/arctic_hydrology.py

And thematically it feels like postprocessing. Consider consolidating and moving to the post-processing module. If nothing else the route is already 1000+ LOC so it could be nice. However, this suggestion is not functional so just take it under consideration.

charparr (approved):
OK, I've gone through the logic in postprocessing.py and the implementation in the routes pretty carefully.  I think we are nailing the desired behavior here...the stabilized ratio `(projected + epsilon) / (historical + epsilon)` is shrinking poorly constrained low-flow change factors toward a value of 1 (i.e. no change) but converging on the established ratio when the baseline is large relative to epsilon.  I think the tests coverage is a nice addition here because we've it covers the blowups we saw before, and the min/mean/max ordering which is an important part of this PR. I did leave some comments that are mostly regarding the assumptions that we are baking into this code...and maybe these should get documented somewhere / ticketed for later consideration if they need to be conveyed to users. The implementation is strong and I have zero issues merging this as is.

cstephen (approved):

